PR #7, opened 2025-08-08
Differentiability Tests

maximilian-gelbrecht:
WIP 

Currently there's still a type instability somewhere, that I need to fix. 

There's also some tooling missing to compute really meaningful gradients that could be compared against finite difference.s


maximilian-gelbrecht:
`energy_to_temperature!` currently has a type instability. I can easily fix that on Monday. But while doing that I noticed that the processes don't have the number type / precision as a parameter. This makes things like this a bit awkward because I want to avoid a `eltype(some_field)` in the code. 

@bgroenks96 What do think about the `NF` has a type parameter for all processes as well? Models already have it.

bgroenks96:
Is it OK if I also make changes to this branch?

maximilian-gelbrecht:
Sure, go ahead

maximilian-gelbrecht:
Great, Thanks Brian! 

I fixed the failing test now.

PR #13, opened 2025-08-11
Replace time stepper in Simulation with timestep cache and add euler step method

### LINKED ISSUE #14: Add generic method for computing forward Euler timesteps
At present, the existing model components re-implement the forward Euler timestep everywhere.

This will change when generic timestepping is implemented, but as an interim solution, @maximilian-gelbrecht suggested to have a method which at least implements the Euler step `progvar = progvar + tendency` to avoid duplicating it.

bgroenks96:
I realized it's stupid to declare the time stepper again in the SImulation (opens up the possibility of inconsistent timestepper definitions between model and simulation).

What we actually want is for `Simulation` to host a cache type for the timestepper, a bit like SciML does. This will be necessary for more complex timesteppers.

I also updated the existing model time stepping implementations to use a common method implementing the forward Euler timestep.

Resolves #14

### DIFF
--- a/src/models/abstract_model.jl
+++ b/src/models/abstract_model.jl
@@ -71,6 +71,11 @@ Returns the time stepping scheme associated with this `model`.
 """
 get_initializer(model::AbstractModel) = model.initializer
 
+"""
+Convenience dispatch for `timestep!` that forwards to `timestep!(state, model, get_time_stepping(model), dt)`.
+"""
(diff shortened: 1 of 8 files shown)


maximilian-gelbrecht (commented):
Yeah, this makes me happy. 

This just needs a tiny unit test for the EulerSolver

PR #18, opened 2025-08-11
Rename project from Terra -> Terrarium

### LINKED ISSUE #16: Rename the package (again) to Terrarium.jl
As pointed out by Sebastian, a subcomponent of ICON-Land is already named TERRA:

https://docs.icon-model.org/land/land.html#ref-land-terra

We don't want to cause any confusion, so we'll rename our model to `Terrarium.jl` instead. This name is preferred by @milankl anyway since it more clearly suggests that the software is a model rather than just a package for processing land/Earth data or something like that.

bgroenks96:
This renames everything, but I think we will need to rename the repository for the doc build and CI to pass.

Resolves #16


maximilian-gelbrecht (approved):
Yes, you need to rename the repo as well then so that the doc build and deploy works again

PR #21, opened 2025-08-12
Improve initialization and add support for input data handling

bgroenks96:
Key changes:

- Remove dependency on CryoGrid (adds adapted implementation of `SoilTexture`); SEB will be refactored into another package.
- WIP


src/processes/atmos/atmosphere.jl
```diff
@@ -0,0 +1,123 @@
...
+    conc = getproperty(state, gas.name)
+    set!(conc, gas.conc)
+end
+
+"""
+Createas a `TracerGas` for ambient CO2 with a prescribed concentration `conc`.
+"""
+AmbientCO2(conc) = TracerGas(:CO2, conc)
+
+@kwdef struct AtmosphericState{
```
milankl, line 30: I like "AtmosphericState" the only confusing I see arising is that it's actually only the atmospheric state in one grid box, not globally? Or is your idea that `::AirTemp` could be a scalar or an array?

src/processes/atmos/atmosphere.jl
```diff
@@ -0,0 +1,123 @@
...
+    T_air::AirTemp = nothing
+
+    "Near-surface specific humidity in kg/kg"
+    humidity::Humidity = nothing
+
+    "Near-surface atmospheric pressure in Pa"
+    pressure::Pressure = nothing
+
+    "Windspeed velocity components in m/s"
+    windspeed::NamedTuple{(:u,:v), Windspeed} = (u=nothing, v=nothing)
```
milankl, line 58: Do you really need `u, v`? I thought wind speed would suffice?

src/processes/atmos/atmosphere.jl
```diff
@@ -0,0 +1,123 @@
...
+
+@kwdef struct TwoBandSolarRadiation{SW, LW} <: AbstractSolarRadiation
+    shortwave::SW = nothing
+    longwave::LW = nothing
+end
+
+variables(::TwoBandSolarRadiation) = (
+    auxiliary(:SwIn, XY(), units=u"W/m^2", desc="Incoming (downwelling) shortwave radiation"),
+    auxiliary(:LwIn, XY(), units=u"W/m^2", desc="Incoming (downwelling) longwave radiation"),
+)
```
milankl, line 118: Your idea is to use shortwave and longwave radiation as forcing but the upwelling fluxes would be computed in the radiation scheme? Then land would need to alter somehow the albedo the atmosphere sees and communicate its skin temperature so that the radiation scheme knows its bottom boundary condition. Sounds good, just asking for clarification

src/processes/atmos/atmosphere.jl
maximilian-gelbrecht: No, it can be an `AbstractArray`, right? 

But we actually would want this to also be flexible enough to take a `RingGrid` here, right?

maximilian-gelbrecht (commented):
This looks good so far! Really cool. 

So, would you still do this as we discussed to put the definition of what the forcing is then in the model type, which then defines a function how to load the inputs/forcing into e.g. the PrecipedFluxes in the auxiliary variables? Or how do you imagine the work flow? Would be great to see at least a rough sketch of the idea.

maximilian-gelbrecht:
What happened? Why did you close and delete the branch?

PR #24, opened 2025-08-21
Refactor boundary condition and initializer types

bgroenks96:
- Add model-specific boundary condition types/constructors
- Add VerticalBoundaryConditions for generically defining top/bottom boundary conditions on any given model
- Generalize flux (forcing) boundary conditions in a PrescribedFlux type; these boundary conditions are currently applied as forcings with boundary_tendency rather than directly as FluxBoundaryConditions. This makes sense for most relevant cases in land modeling, I think.
- Add boundary condition type for atmospheric state variables


Project.toml
```diff
@@ -12,11 +12,13 @@ Dates = "ade2ca70-3891-5945-98fb-dc099432e06a"
 DocStringExtensions = "ffbed154-4ef7-542d-bbb7-c09d3a79fcae"
 Flatten = "4c728ea3-d9ee-5c9a-9642-b6f7d7dc04fa"
 FreezeCurves = "71e4ad71-e4f2-45a3-aa0a-91ffaa9676be"
+Interpolations = "a98d9a8b-a2ab-59e6-89dd-64a1c18fca59"
 JLD2 = "033835bb-8acc-5ee8-8aae-3f567f8a3819"
 KernelAbstractions = "63c18a36-062a-441e-b654-da1e3ab1ce7c"
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
 ModelParameters = "4744a3fa-6c31-4707-899e-a3298e4618ad"
 NCDatasets = "85f8d34a-cbdd-5861-8df4-14fed0d494ab"
+NamedTupleTools = "d9ec5142-1e00-5aa0-9d6a-321866360f50"
```
maximilian-gelbrecht, line 21: The package is not very well maintentained and has some type piracy issues, if you use only one function from it, it might be worth it to just reimplement that function.

src/boundary_conditions.jl
```diff
@@ -0,0 +1,141 @@
...
+Updates any state variables associated with the given boundary conditions.
+"""
+compute_auxiliary!(state, model, ::AbstractBoundaryConditions) = nothing
+compute_auxiliary!(state, model, ::BoundaryCondition) = nothing
+compute_auxiliary!(state, model, ::VarBoundaryConditions) = nothing
+
+"""
+    $SIGNATURES
+
+Computes the boudnary tendency for the grid cell at `loc`; zero for all other grid cells.
```
maximilian-gelbrecht, line 38: typo

src/initializers.jl
```diff
@@ -0,0 +1,50 @@
+# Initializer interface
```
maximilian-gelbrecht, line 1: I know there's already an explanation in the general docs, but I think the initialiser interface could use some more explanation in doc strings as well somewhere. Especially with regards to how this is used in the end in the example scripts, which isn't immediately clear I think from the docs as is.

test/atmos/atmosphere.jl
```diff
@@ -0,0 +1,42 @@
...
+    pressure = 1.0
+    windspeed = 1.0
+    tracers = TracerGases(AmbientCO2(1.0))
+    atm = AtmosphericState(; grid, precip, solar, T_air, humidity, pressure, windspeed, tracers)
+    @test haskey(atm.tracers, :CO2)
+    atm_varnames = map(Terrarium.varname, variables(atm))
+    @test all(map(∈(atm_varnames), (:T_air, :humidity, :pressure, :windspeed, :rainfall, :snowfall, :SwIn, :LwIn)))
+    # check compute_auxiliary!
+    state = (
+        T_air = Field{Center,Center,Nothing}(get_field_grid(grid)),
```
maximilian-gelbrecht, line 22: We'd always use a Oceanigans.Field here? 

Then, we would interpolate between the RingGrids.Field and the Oceanigans.Field before?

maximilian-gelbrecht (commented):
It's still a bit unclear to me how one would use the `AtmosphericState` in the end as a forcing, could you briefly sketch that? Doesn't have to fully implemented just the idea that you currently have for that.

maximilian-gelbrecht:
> The basic gist of it is something like
> 
> ```julia
> atmos = AmtmosphericState(; T_air=T_air_input_data, ...)
> seb = SurfaceEnergyBalance(; atmos, other_args...)
> soil_bc = SoilBoundaryConditions(top=seb)
> soil_model = SoilModel(; boundary_conditions=soil_bc)
> sim = initialize(soil_model; inputs)
> ```
> 
> and then the `SurfaceEnergyBalance` code could simply reference `state.T_air` which would be updated at each timesetep in `compute_axuiliary!`.
> 
> The problem I am still grappling with is where exactly the input data should live. I don't like the current concept of putting it directly into the model. That would mean, for example, that _all_ of the input data would need to be copied onto the GPU.

I think I would, in very rough terms, just define a 


```Julia 
struct Forcing 
       parameters # some parameters the forcing might have 
end 

function get_forcing_data(f::Forcing, clock::Clock) 
       # get the data from wherever it is 
end 
``` 

`Forcing` (or call it `InputData`)  would be a field of the model then, and in same cases (e.g. for a constant forcing) one might save it as a field in `Forcing`, for some cases it might open a netcdf file, for some cases it might get it from another model, etc.

maximilian-gelbrecht:
Putting it in the `Simulation` is a good idea, yes! And I am indifferent towards how exactly it's called. If it's not forcing, that's also fine.

maximilian-gelbrecht:
I am probably only able to do a review early next week.

bgroenks96:
It took an unreasonably long time, but I think I finally have something here that I am relatively happy with.

The basic idea is that input variables are treated like another state variable type, e.g:

```
variables(process::SomeProcess) = (
    prognostic(:progvar, XY()),
    auxiliary(:auxvar, XY()),
    input(:forcingvar, XY()),
    ...
)
```

However, unlike the other state variable types, input `Field`s are allocated and stored in an external structure `InputFields` which is supplied to `StateVariables`.

Currently, `InputFields` is provided by another type called `InputProvider` which acts just as a container and interface between the `InputFields` (memory) and one or more `AbstractInputSource`s (loaders) which can have any number of implementations. The `InputProvider` is stored in the `Simulation` thus decoupling the input loading completely from the model. This also has the advantage of allowing the inputs to be easily manipulated directly if necessary, including differentiating them with Enzyme, I think (not yet tested though).

I still need to make a few more changes updating the existing types to actually use the new `InputVariable`s but otherwise I think it's in a good place. One thing we should decide is whether or not we still need `AtmosphericState`; the input variable could just be declared directly by the model components as is currently done in the `VegetationModel` code. Let me know what you think @maximilian-gelbrecht and maybe also @mahabadri.

I would also like to note that I'm not totally happy with the names for the various input types so feel free to suggest alternatives.

bgroenks96:
@maximilian-gelbrecht Looks like using `fastiterate` from CryoGrid.jl solved the Enzyme issue.

src/inputs/input_provider.jl
```diff
@@ -0,0 +1,26 @@
+"""
+    $TYPEDEF
+
+Acts as an interface between a given `InputFields` container and one or more `AbstractInputSource`s
+that are responsible for intiializing and updating the input `Field`s at each time step.
```
maximilian-gelbrecht, line 5: typo

src/inputs/input_provider.jl
```diff
@@ -0,0 +1,26 @@
...
+
+    "A list of input sources to be invoked by `update_inputs!`"
+    sources::Vector{SourceType}
+end
+
+InputProvider(grid::AbstractLandGrid, sources::AbstractInputSource...) = InputProvider(InputFields(grid), collect(sources))
+
+function update_inputs!(provider::InputProvider, clock::Clock)
+    for source in provider.sources
+        update_inputs!(provider.fields, source, clock)
```
maximilian-gelbrecht, line 24: I was thinking back and forth a bit, I think I am good with a container format here to interface between the input sources and the variables, especially for situations in which we have more than one input source. We should probably have a convience constructor for this either here or in the Simulation though so that users don't have to do this in most simulation and can also only hand over a `InputSource`

test/inputs/inputs.jl
```diff
@@ -0,0 +1,86 @@
...
+    grid = ColumnGrid(ExponentialSpacing())
+    X1 = Field(grid, XY())
+    set!(X1, 1)
+    field_input = FieldInputSource(; X1)
+    provider = InputProvider(grid, field_input)
+    clock = Clock(time=0)
+    update_inputs!(provider, clock)
+    @test get_input_field(provider.fields, :X1, XY()) === X1
+    @test all(get_input_field(provider.fields, :X1, XY()) .== 1)
+end
```
maximilian-gelbrecht, line 86: Could we define just a very basic forcing test in which we just add a constant field to some variable in the state variable at every time step?

src/processes/prescribed_atmosphere.jl
```diff
@@ -0,0 +1,81 @@
...
+
+    "Downwelling solar radiation scheme"
+    solar::SolarRad = TwoBandSolarRadiation()
+
+    "Atmospheric tracer gases"
+    tracers::Tracers = TracerGases(AmbientCO2())
+end
+
+variables(atm::PrescribedAtmosphere) = (
+    input(:T_air, XY(), units=u"°C", desc="Near-surface air temperature in °C"),
```
maximilian-gelbrecht, line 59: I like that, as we just discussed. Having a predefined atmosphere struct here based on the standard input interface is good.

bgroenks96:
@maximilian-gelbrecht Is there anything still needed here?

maximilian-gelbrecht:
I'll have a final look at the PR tomorrow!

maximilian-gelbrecht (approved):

Issue #10, opened 2025-08-11
Improve differentiability test


maximilian-gelbrecht:
#7 implemented basic differentiability, but currently we get `NaN` gradients in these tests. 

- [ ] Investigate and fix the `NaN` gradients
- [ ] Implement finite differences comparison

PR #25, opened 2025-08-22
Track down differentiability issues 

maximilian-gelbrecht:
I started to track down the source of the NaNs #10 

`compute_auxiliary!` and `compute_tendencies!` are fine. I pretty sure the problem is caused by `energy_to_temperature!` not sure yet why. I will probably have to do some trial and error.


maximilian-gelbrecht:
Ok, so my investigation first showed that the NaN derivatives enter via `pore_water_ice_saturation` and `internal_energy`, which points to the `ifelse` clause there: https://github.com/TUM-PIK-ESM/Terrarium.jl/blob/d956a7e006aba7d7b101000d9c75befd95294c94/src/processes/soil/soil_energy.jl#L222

Replacing that with just `state.liquid_water_fraction[i, j, k] = zero(sat)` for a test also yields finite gradients for the complete model, so this is the likely culprit. 

I have to stop working on that for now, and will come back to this with a proper fix later. 

Small side node: the `StateVariables` are already compatible with `FiniteDifferences` as they are, so we won't need to write an extension for that.

bgroenks96:
Hmm, maybe the other `ifelse` needs to be replaced as well? What happens if you just set `T = U / C` and comment out that if/else?

bgroenks96:
@maximilian-gelbrecht The `NaN` gradients are fixed now. It was caused by the division-by-zero in the phase change branch of the liquid water content calculation in cases where the porosity (and thus latent heat content) was zero. The porosity is currently zero by default unless you explicitly initialize it.

bgroenks96:
@maximilian-gelbrecht I have replaced the `Average` operator with a direct evaluation of `mean(interior(state.temperature))` and this works. I would suggest that we go ahead and merge this PR once the CI passes.

PR #33, opened 2025-10-01
Add land-sea mask to global grid and improve forcing data handling

bgroenks96:
This PR consists of three key changes:

- `GlobalRingGrid` is now renamed to `ColumnRingGrid` (to more accurately represent its relation to `ColumnGrid`) and a new `mask` field is included to allow non-land grid points to be efficiently masked out
- The interface for specifying boundary condition is (hopefully) simplified and directly integrated with `Oceananigans` `BoundaryCondition` types. A new `Input` type is also included to act as a placeholder for input variables.
- The global heat conduction example script has been updated to read in ERA5-Land air temperature data and apply it as a Dirichlet-type upper boundary condition to force the soil energy balance.

~~Currently, the global simulation is becoming unstable (`NaN` prognostic variables) after a few days of simulated time, and it's not yet clear why this is. I need to check the forcing data and make sure there aren't some sneaky `NaN`s in there...~~

Turns out the NaNs were in fact due to issues with the forcing, not instability.

### DIFF
--- a/ext/TerrariumRastersExt/TerrariumRastersExt.jl
+++ b/ext/TerrariumRastersExt/TerrariumRastersExt.jl
@@ -4,6 +4,7 @@ using Terrarium
 using Terrarium: XY, XYZ # variable dims
 
 using Dates
+using DocStringExtensions
 using Interpolations
 using Oceananigans
 using Rasters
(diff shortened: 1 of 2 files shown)


bgroenks96:
@maximilian-gelbrecht I think this PR is ready to merge unless you have any other concerns?

Issue #27, opened 2025-09-02
Default model constructors and number precision


maximilian-gelbrecht:
So, currently we have model constructors like this: 

https://github.com/TUM-PIK-ESM/Terrarium.jl/blob/d956a7e006aba7d7b101000d9c75befd95294c94/src/models/soil_model.jl#L9-L47 

As far as I can see it this doesn't make sure that the same number precision is used across the model when the default field is constructed. This should probably be changed across all models, right?

maximilian-gelbrecht:
Okay, I can do that.

maximilian-gelbrecht:
Before I do all of that. In Speedy we have the `SpectralGrid` that is used in all component initialisation. For the atmospheric model that's quite obvious, it sets the discritzation, but it also sets these computational parameters like the precision. 

For Terrarium, we don't need the discretisation in every model process and component, but do we potentially need more common model setup parameters than just the precision? Then we should probably introduce a type for that. 

If it's really only the precision then we just do 
```
@kwdef ModelComp{NF}
        process::Process = DefaultProcess{NF}()
```

bgroenks96:
```
@kwdef struct ModelComp{NF}
        process::Process = DefaultProcess{NF}()
end 
```

I don't think that works? I get an error that `NF` is not defined.

I think that's stupid and should be considered a bug in Julia but...

One obvious candidate for a universal field would be the `PhysicalConstants` struct.  We could also define some other `Config` type (which itself could hold the constants). That might be more future-proof.

maximilian-gelbrecht:
Ok, yeah `Config`, holding `PhysicalConstants` as well sounds reasonable to me. I can make a draft of that as soon as I find some time for it.

milankl:
Don't you need the struct then to be parametric wrt every field? That sounds unnecessary?

bgroenks96:
Nope! That's the beauty of Flatten.jl :)

You give it a type like AbstractFloat and then it only selects the fields in the nested structure with that type. You can also exclude certain fields or types with `Flatten.flattenable`.

milankl:
Reconstruction things sounds okay to me but also not ideal because I'd be worried that people hacking along in the model end up in weird situations that some components are reconstructed others aren't and it wouldn't be very obvious what's going on?

milankl:
[...]

PR #34, opened 2025-10-01
Add generalized timestepping w/ time step adaptation

bgroenks96:
WIP


bgroenks96:
@maximilian-gelbrecht This PR is ready for review

examples/soil_heat_column.jl
maximilian-gelbrecht: Not directly related to this PR, but what do you think about a seperate CI workflow to check that the examples are running? Imho, outdated examples could really become a problem. It certainly is in many other projects. The only problem with that are GPU examples, but one could just add a `if CUDA.functional()` to the script.

src/grids/column_grid.jl
```diff
@@ -1,10 +1,10 @@
 """
-    ColumnGrid{NF,RectGrid<:OceananigansGrids.RectilinearGrid} <: AbstractLandGrid
+    ColumnGrid{NF, Arch<:AbstractArchitecture, RectGrid<:OceananigansGrids.RectilinearGrid} <: AbstractLandGrid
 
 Represents a set of laterally independent vertical columns with dimensions (x, y, z)
 where `x` is the column dimension, `y=1` is constant, and `z` is the vertical axis.
 """
-struct ColumnGrid{NF,RectGrid<:OceananigansGrids.RectilinearGrid} <: AbstractLandGrid{NF}
+struct ColumnGrid{NF, Arch, RectGrid<:OceananigansGrids.RectilinearGrid} <: AbstractLandGrid{NF, Arch}
```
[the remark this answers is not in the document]
maximilian-gelbrecht: About the `Architecture` do you want to have it as a parameter to dispatch on it, or why is it needed?

src/model_state.jl
```diff
@@ -0,0 +1,150 @@
+"""
+    $TYPEDEF
+
+Represents the state of a "simulation" for a given `model`. `ModelState` consists of a
+`clock`, a `model`, and an initialized `StateVariables` data structure, as well as a `cache`
+for the timestepper and any relevant `inputs` provided by a corresponding `InputProvider`.
+The `ModelState` implements the `Oceananigans.AbstractModel` interface and can thus be
+treated as a "model" in `Oceananigans` `Simulation`s and output reading/writing utilities.
+"""
+struct ModelState{
```
maximilian-gelbrecht: I already wrote it and told to you, I am fine with this. I think this is a good and pragmatic compromise, and I don't see an immediate downside.

src/model_state.jl
```diff
@@ -0,0 +1,150 @@
...
+time_step!(state::ModelState, Δt; kwargs...) = timestep!(state, Δt)
+
+"""
+    $TYPEDEF
+
+Resets the simulation `clock` and calls `initialize!(state, model)` on the underlying model which
+should reset all state variables to their values as defiend by the model initializer.
+"""
+function initialize!(state::ModelState)
+    # TODO: reset other variables too?
```
maximilian-gelbrecht: I am a bit confused, does the `initialize!`(::State, ::Model)` reset the `State` variables? I thought not. Because the doctoring is kind of implying that, but here you are wondering whether they should be reset separately.

src/models/soil/soil_model.jl
```diff
@@ -17,33 +17,33 @@ $(TYPEDFIELDS)
     BoundaryConditions<:AbstractBoundaryConditions,
     Initializer<:AbstractInitializer,
     TimeStepper<:AbstractTimeStepper,
-} <: AbstractSoilModel{NF}
+} <: AbstractSoilModel{NF, GridType, TimeStepper}
```
maximilian-gelbrecht: We still have this discussion in #27, do we settle this for now with explicitly handing over the precision?

examples/soil_heat_column.jl
```diff
@@ -21,9 +22,9 @@ sim = initialize(model)
 # run simulation forward for a set period of time
 run!(sim, period=Day(10))
```
maximilian-gelbrecht: `modelstate` instead of `sim`?

src/model_state.jl
maximilian-gelbrecht: Just that I understand correctly. Currently we don't have a `Simulation` anymore, but we would probably also reuse or subtype Oceananigans' simulation to get the output to work, correct?

maximilian-gelbrecht (commented):
Looks great!

examples/soil_heat_column.jl
maximilian-gelbrecht: Yeah, I am happy to set this up tomorrow or so.

src/grids/column_grid.jl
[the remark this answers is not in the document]
bgroenks96: It's because the Oceananigans `AbstractModel`  currently requires this as a type parameter (this will likely change in Oceananigans in the near future).

src/model_state.jl
bgroenks96: Yes, we can re-use Oceananigans `Simulation` now. Running it works fine, but output isn't working yet because of some issues with the grid.

src/timesteppers/abstract_timestepper.jl
```diff
@@ -1,29 +1,95 @@
...
-function get_dt end
+function default_dt end
 
 """
     is_adaptive(timestepper::AbstractTimeStepper)
 
-Returns `true` if the given time stepper is adaptive, false otherwise.
+Return `true` if the given time stepper is adaptive, false otherwise.
 """
 function is_adaptive end
```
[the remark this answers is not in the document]
bgroenks96: I would suggest adding this once we add implicit solvers.

src/timesteppers/forward_euler.jl
```diff
@@ -3,20 +3,25 @@ Simple forward Euler time stepping scheme.
...
 
-@inline step(::ForwardEuler, progvar, tendency, dt) = progvar + dt*tendency
-
-initialize(::AbstractModel{NF}, ::ForwardEuler) where {NF} = ForwardEulerCache{NF}()
-
-# function timestep!(state, model::AbstractModel, euler::ForwardEuler, dt=get_dt(euler))
-#     # TODO: implement timestep! generically
-#     error("not implemented")
-# end
+function timestep!(state, model::AbstractModel, timestepper::ForwardEuler, Δt = default_dt(timestepper))
```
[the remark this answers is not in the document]
bgroenks96: regarding `AbstractExplicitTimeStepper`, I actually did that but then reverted it for now to keep things as simple as possible until we actually add implicit schemes.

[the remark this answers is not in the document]
bgroenks96: In your improved step, you're still using the original state at time $t$ which seems redundant? Wouldn't most RK/multi-stage methods first do the `explicit_step!`? For example:

```julia
stage = copy(state) 

# initial step
compute_auxiliary!(stage, model) 
compute_tendencies!(stage, model)
explicit_step!(stage, model, Midpoint(), dt/2)

# final update
compute_auxiliary!(stage, model) 
compute_tendencies!(stage, model)
set_tendencies!(state, stage)
explicit_step(state, Midpoint(), model, dt)  
```

src/timesteppers/forward_euler.jl
```diff
@@ -3,20 +3,25 @@ Simple forward Euler time stepping scheme.
 """
```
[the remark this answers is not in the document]
bgroenks96: [53b792a](https://github.com/TUM-PIK-ESM/Terrarium.jl/pull/34/commits/53b792a7b642c9df8225be20ee8209094fdd21ca)

src/model_state.jl
maximilian-gelbrecht: Am I misunderstanding this, or is this a rather big change, why would you compute the auxiliary variables before the timestep?

bgroenks96:
@maximilian-gelbrecht Ok there were some unexpected issues that came up but should be ready to merge now.

PR #37, opened 2025-10-15
Add initial implementation of Richards' equation for soil water fluxes

### LINKED ISSUE #31: Implement soil water flow via Richards equation
As currently described [in the docs](https://tum-pik-esm.github.io/Terrarium.jl/dev/physics/soil_energy_water/#Vertical-water-transport-in-variably-saturated-soil).

bgroenks96:
The goal of this PR is to add an initial implementation of vertical water flow in soils via the conservation law:

```math
\phi(z) \frac{\partial s_{\text{wi}}(\Psi(z,t))}{\partial t} = \nabla \cdot K(s_{\text{wi}}, T) \nabla \Psi
```
where $s_{\text{wi}}$ is the saturation level of water and ice in the soil pores, $\phi(z)$ is the (spatially variable) porosity, $K$ is the hydraulic conductivity, $T$ is temperature, and $\Psi = \psi_m + \psi_h + \psi_z$ is the total hydraulic pressure head which is the sum of the matric potential, hydrostatic pressure, and elevation head.

In the unsaturated (vadose) zone of the soil where hydrostatic pressure is absent, this reduces to the well known Richardson-Richards' equation:

```math
\phi(z) \frac{\partial s_{\text{wi}}(\Psi(z,t))}{\partial t} = \nabla \cdot K(s_{\text{wi}}, T) \nabla \left(\psi_m + 1 \right)
```

This describes the vertical movement of water according to gravity-driven percolation and capillary-driven diffusion.

Currently, the Richards' equation is implemented as a specific choice of operator on the `SoilHydrology` type.

The unsaturated hydraulic conductivity and soil water retention relation (i.e. the bijective mapping $f: \psi_m \leftrightarrow s_{\text{wi}}$) is specified in the `SoilHydraulics` and `UnsatK` types.

Coupling of the matric potential to freeze-thaw via Clausius-Clapeyron is not yet implemented and will come in a future PR.

Additional changes:
- Adds a new type `SoilComposition` to manage the relationship between characteristic/saturation fractions of each material constituent and their corresponding scaled volumetric fractions.
- Adds a new dispatch of `adapt` for scalar types that uses `Flatten.reconstruct`  to replace all numeric types. This was necessary (at least for now) to align the numeric types from `FreezeCurves`. I am also in favor of adopting this approach more broadly in Terrarium to ensure type consistency.

TODO:
- [x] Handle oversaturation or undersaturation of grid cells via timestepping restrictions or instantaneous redistribution
- [x] Add differentiability tests
- [x] Check if RRE implementation works on GPU

Resolves #31

@mahabadri @maximilian-gelbrecht Happy to get your feedback.

### DIFF
--- a/test/differentiability/soil_hydrology_diff.jl
+++ b/test/differentiability/soil_hydrology_diff.jl
@@ -81,7 +81,7 @@ end
     @test all(isfinite.(dstate.temperature))
 end
 
-@testset "Soil hydrology: compute_auxiliary! RRE" begin
+@testset "Soil hydrology: compute_tendencies! RRE" begin
     hydraulic_properties = SoilHydraulicsSURFEX(Float64)
     model = build_soil_energy_hydrology_model(CPU(), Float64; hydraulic_properties)
     model_state = initialize(model)


src/processes/soil/hydrology/soil_hydrology_rre.jl
```diff
@@ -0,0 +1,207 @@
...
+@inline function darcy_flux(i, j, k, grid, state)
+    # Get pressure head
+    ψ = state.pressure_head
+    # Get hydraulic conductivity
+    K = state.hydraulic_conductivity
+    # Darcy's law: q = -K ∂ψ/∂z
+    # TODO: also account for effect of temperature on matric potential
+    ∇ψ = ∂zᵃᵃᶠ(i, j, k, grid, ψ)
+    ## Take minimum of hydraulic conductivities in the direction of flow
+    Kₖ = (∇ψ < 0)*min(K[i, j, k-1], K[i, j, k]) +
```
bgroenks96: @glwagner I would appreciate your feedback on this approach. Maybe you have an idea for how this can be done better. Averaging using the interpolation operator is not viable for hydraulic conductivity because (1) it varies substantially (by order of magnitude) over space and (2) it is not physically justified. Taking the minimum of the hydraulic conductivities is a fairly common choice in hydrological modeling based on the physical intuition that the flow path of the water through the pore space is going to be limited by the most restrictive (least conductive) material along that path. Another option would be the harmonic mean.

However, I am using the minimum along the direction of flow here, which I am not sure about.

glwagner: I am not sure if I understand what you want, but you can define a function:

```julia
minᵃᵃᶠ(i, j, k, grid, a) = @inbounds min(a[i, j, k], a[i, j, k-1])
```

I don't agree that averaging is not physically justified. It is _mathematically_ justified (it is part of the staggered formulation). This is widely used in C-grid LES models, for example (not just Oceananigans --- all C-grid LES models for atmospheres and oceans). You can come up with another formulation for sure, but it is not correct to argue that averaging is not physical justified.

bgroenks96: It is indeed not physically justified because soil properties (e.g. saturation) are not continuous and hydraulic conductivities can vary by orders of magnitude between soil layers. Averaging K across cells is well known to produce wildly unrealistic soil water flow.

Take for example a very straightforward case of a frozen soil volume below a thawed soil volume. In the frozen volume, the conductivity is near zero since liquid water is immobile, but in the thawed volume, it is nonzero. Averaging would result in a nonzero flux into the frozen soil volume which is physically incorrect.

glwagner: Presumably you have a model for conductivity. In this model, you will interpolate the inputs of the model to cell interfaces, and then use the model to compute conductivity there.

src/processes/soil/hydrology/soil_hydrology_rre.jl
```diff
@@ -0,0 +1,309 @@
...
+    hydrology::SoilHydrology,
+    strat::AbstractStratigraphy,
+    bgc::AbstractSoilBiogeochemistry,
+)
+    i, j, k = @index(Global, NTuple)
+    soil = soil_composition((i, j, k), state, strat, hydrology, bgc)
+    state.hydraulic_conductivity[i, j, k] = hydraulic_conductivity(hydrology.hydraulic_properties, soil)
+    # Manually set hydraulic conductivity at the boundaries to the conductivity of the cell centers at the boundaries;
+    # note that this is an assumption that could potentially be relaxed in the future.
+    # TODO: maybe this should be done with boundary conditions?
```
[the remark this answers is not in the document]
matteo-willeit: I agree that hydraulic conductivity should be defined on grid cell faces.

maximilian-gelbrecht (approved):

PR #39, opened 2025-10-21
Add initial implementation of surface energy balance

### LINKED ISSUE #40: Implement initial surface energy balance scheme with coupling to bare soil and vegetation
Implement the surfarce energy balance for bare soil and vegetated surfaces following the formulations of PALADYN (Willeit 2016) and CryoGrid (Westermann et al. 2023). The SEB is broadly represented as:

```math
0 = S_{\downarrow} - S_{\uparrow} + L_{\downarrow} - L_{\uparrow}(T_0) + H_s(T_0) + H_L(T_0) + G(T_0)
```
where $T_0$ is the skin temperature, $S$ and $L$ are the shortwave and longwave radiation with arrows indicating the direction (incoming or outgoing), $H_s$ is the sensible heatflux, $H_l$ is the latent heat flux, and $G$ is the ground heat flux.

These terms can also be further partitioned to represent the contributions of different vertical levels of the canopy, but this is outside of the scope of this PR (and possibly Terrarium generally for the time being).

This issue can be considered resolved once the following are implemented:
- Prescribed and diagnosed radiative fluxes (with constant albedo and emissivity) 
- Prescribed and diagnosed turbulent fluxes (with a dummy implementation of aerodynamic resistance)
- Prescribed and prognostic skin temperature with corresponding ground heat flux
- Basic coupling to bare soil and a simple "big leaf" vegetation scheme

bgroenks96:
This PR provides an initial implementation of the surface energy balance, including:
- Albedo/emissivity
- Radiative budget (short- and longwave radiation in and out)
- Turbulent fluxes (sensible and latent)
- Skin temperature

Since all of these variables are commonly available in gridded data products, their process implementations all include both `Prescribed` and `Diagnosed` variants; the former simply define `input` variables (and in some cases diagnosed downstream quantities like net radiation) while the latter perform the actual SEB calculations based on atmospheric forcings (incoming radiation, air temperature, pressure, humidity, and wind speed).

The core SEB equation (under the positive flux upward convention) is:

```math
R_{\text{net}} = S_{\uparrow} - S_{\downarrow}  + L_{\uparrow}(T_0) - L_{\downarrow} = H_s(T_0) + H_L(T_0) + G(T_0)
```
where $T_0$ is the skin temperature, $S$ and $L$ are the shortwave and longwave radiation with arrows indicating the direction (incoming or outgoing), $H_s$ and $H_l$ are the sensible and latent heat fluxes respectively, and $G$ is the ground heat flux.

The main objective of the SEB is to solve for the skin temperature (as well as other unknowns depending on the formulation) such that the above equation is satisfied.

The current implementation of `ImplicitSkinTemperature` does this following a four step strategy:

1. Compute all fluxes using $T_0 = T_1$ where $T_1$ is the temperature of the uppermost ground layer. This is effectively an initial guess of the skin temperature.
2. Compute $G = R_{\text{net}} - H_s -H_l$
3. Solve for $T_0 = T_s + \kappa_1 G / (2 \Delta z_1)$
4. Recompute all fluxes with the diagnosed $T_0$

This basically constitutes a single step of a fixed point iteration that *should* converge to the true skin temperature. As such, there is always some error in the estimate, with the magnitude of that error increasing for larger fluctuations in the radiative forcing.

Resolves #40

### DIFF
--- a/src/processes/abstract_types.jl
+++ b/src/processes/abstract_types.jl
@@ -84,11 +84,11 @@ Base type for radiation budget schemes.
 abstract type AbstractRadiativeFluxes <: AbstractProcess end
 
 """
-    net_incoming_radiation(idx, state, ::AbstractRadiativeFluxes)
+    surface_net_radiation(idx, state, ::AbstractRadiativeFluxes)
 
 Return the current net radiation at the given `idx`.
 """
-function net_incoming_radiation end
+function surface_net_radiation end
 
 """
 Base type for turbulent (latent and sensible) heat fluxes at the surface.
--- a/src/processes/surface_energy/radiative_fluxes.jl
+++ b/src/processes/surface_energy/radiative_fluxes.jl
@@ -1,7 +1,7 @@
 """
-Return the current value of the `net_incoming_radiation` at the given `idx`.
+Return the current value of the `surface_net_radiation` at the given `idx`.
 """
-@inline net_incoming_radiation(idx, state, ::AbstractRadiativeFluxes) = state.net_incoming_radiation[idx...]
+@inline surface_net_radiation(idx, state, ::AbstractRadiativeFluxes) = state.surface_net_radiation[idx...]
 
 """
     $TYPEDEF
@@ -18,7 +18,7 @@ struct PrescribedRadiativeFluxes <: AbstractRadiativeFluxes end
 variables(::PrescribedRadiativeFluxes) = (
     input(:surface_shortwave_up, XY(), units=u"W/m^2", desc="Outoing (upwelling) shortwave radiation"),
     input(:surface_longwave_up, XY(), units=u"W/m^2", desc="Outoing (upwelling) longwave radiation"),
-    auxiliary(:net_incoming_radiation, XY(), units=u"W/m^2", desc="Net radiation budget"),
+    auxiliary(:surface_net_radiation, XY(), units=u"W/m^2", desc="Net outgoing (positive up) radiation"),
 )
 
 function compute_auxiliary!(state, model, rad::PrescribedRadiativeFluxes)
@@ -37,7 +37,7 @@ struct DiagnosedRadiativeFluxes <: AbstractRadiativeFluxes end
 variables(::DiagnosedRadiativeFluxes) = (
     auxiliary(:surface_shortwave_up, XY(), units=u"W/m^2", desc="Outoing (upwelling) shortwave radiation"),
     auxiliary(:surface_longwave_up, XY(), units=u"W/m^2", desc="Outoing (upwelling) longwave radiation"),
-    auxiliary(:net_incoming_radiation, XY(), units=u"W/m^2", desc="Net radiation budget"),
+    auxiliary(:surface_net_radiation, XY(), units=u"W/m^2", desc="Net radiation budget"),
 )
 
 function compute_auxiliary!(state, model, rad::DiagnosedRadiativeFluxes)
@@ -81,7 +81,7 @@ end
     # compute outputs
     state.surface_shortwave_up[i, j, 1] = surface_shortwave_up = shortwave_out(rad, surface_shortwave_down, α)
     state.surface_longwave_up[i, j, 1] = surface_longwave_up = longwave_out(rad, consts, surface_longwave_down, Tsurf, ϵ)
-    state.net_incoming_radiation[i, j, 1] = net_incoming_radiation(rad, surface_shortwave_down, surface_shortwave_up, surface_longwave_down, surface_longwave_up)
+    state.surface_net_radiation[i, j, 1] = surface_net_radiation(rad, surface_shortwave_down, surface_shortwave_up, surface_longwave_down, surface_longwave_up)
 end
 
 @kernel function compute_net_radiation!(state, ::AbstractLandGrid, rad::AbstractRadiativeFluxes, atmos::AbstractAtmosphere)
@@ -94,20 +94,20 @@ end
     surface_longwave_down = longwave_in(idx, state, atmos)
     
     # compute net radiation
-    state.net_incoming_radiation[i, j, 1] = net_incoming_radiation(rad, surface_shortwave_down, surface_shortwave_up, surface_longwave_down, surface_longwave_up)
+    state.surface_net_radiation[i, j, 1] = surface_net_radiation(rad, surface_shortwave_down, surface_shortwave_up, surface_longwave_down, surface_longwave_up)
 end
 
 # Kernel functions
 
 """
-    net_incoming_radiation(::AbstractRadiativeFluxes, surface_shortwave_down, surface_shortwave_up, surface_longwave_down, surface_longwave_up)
+    surface_net_radiation(::AbstractRadiativeFluxes, surface_shortwave_down, surface_shortwave_up, surface_longwave_down, surface_longwave_up)
 
 Compute the net radiation budget given incoming and outgoing shortwave and longwave radiation.
 """
-@inline function net_incoming_radiation(::AbstractRadiativeFluxes, surface_shortwave_down, surface_shortwave_up, surface_longwave_down, surface_longwave_up)
+@inline function surface_net_radiation(::AbstractRadiativeFluxes, surface_shortwave_down, surface_shortwave_up, surface_longwave_down, surface_longwave_up)
     # Sum up radiation fluxes; note that by convention fluxes are positive upward
-    net_incoming_radiation = surface_shortwave_up - surface_shortwave_down + surface_longwave_up - surface_longwave_down
-    return net_incoming_radiation
+    surface_net_radiation = surface_shortwave_up - surface_shortwave_down + surface_longwave_up - surface_longwave_down
+    return surface_net_radiation
 end
 
 """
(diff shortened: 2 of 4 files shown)


src/processes/abstract_types.jl
```diff
@@ -14,14 +14,125 @@ variables(op::AbstractOperator) = ()
...
+
+# Surface energy balance types
+
+"""
+Base type for surface albedo and emissivity process implementations.
+"""
+abstract type AbstractAlbedo <: AbstractProcess end
+
+"""
+    alebdo(idx, state, ::AbstractAlbedo)
```
maximilian-gelbrecht: typo

src/processes/abstract_types.jl
```diff
@@ -14,14 +14,125 @@ variables(op::AbstractOperator) = ()
...
+
+Return the current albedo at the given `idx`.
+"""
+function albedo end
+
+"""
+    emissivity(idx, state, ::AbstractAlbedo)
+AbstractAlbedo
```
maximilian-gelbrecht: typo

src/processes/processes.jl
```diff
@@ -8,17 +10,41 @@ include("physics_utils.jl")
...
+include("soil/biogeochem/soil_texture.jl")
 
-# Vegetation
+export SoilComposition, volumetric_fractions
+include("soil/biogeochem/soil_composition.jl")
+
+export HomogeneousSoil
+include("soil/biogeochem/homogeneous_soil.jl")
+
+export ConstantSoilCarbonDenisty
```
maximilian-gelbrecht: typo

src/processes/surface_energy/albedo.jl
maximilian-gelbrecht: A bit more general question here: 

I am still a bit confused about how different model components couple. So here in the albedo, in the long term, we'd probably compute the albedo also based on soil properties and vegetation. Would this be handles just by directly diagnosing from the state variables from the other components, or would you imagine a different structure for that.

src/abstract_model.jl
```diff
@@ -77,6 +77,28 @@ Convenience dispatch for `timestep!` that forwards to `timestep!(state, model, g
...
 timestep!(state, model::AbstractModel, Δt) = timestep!(state, model, get_time_stepping(model), Δt)
 
+"""
+Convenience constructor for all `AbstractLandModel` types that allows the `grid` to be passed
+as the first positional argument.
+"""
+(::Type{Model})(grid::AbstractLandGrid; kwargs...) where {Model<:AbstractModel} = Model(; grid, kwargs...)
+
+function Adapt.adapt_structure(to, model::AbstractModel)
+    return setproperties(model, map(prop -> Adapt.adapt(to, prop), getproperties(model)))
```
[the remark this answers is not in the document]
bgroenks96: I don't remember this being part of this PR, but I think the point was to allow for a default `adapt_structure` implementation for all model types where we don't necessarily know the concrete fields or constructors from the abstract type.

It's not clear if we actually need this actually? I'll try deleting it and see if anything breaks.

src/grids/grid_utils.jl
```diff
@@ -24,6 +29,13 @@ const RingGridOrField = Union{RingGrids.AbstractGrid, RingGrids.AbstractField}
 on_architecture(::GPU, obj::RingGridOrField) = RingGrids.Architectures.on_architecture(RingGrids.Architectures.GPU(), obj)
 on_architecture(::CPU, obj::RingGridOrField) = RingGrids.Architectures.on_architecture(RingGrids.Architectures.CPU(), obj)
 
+# Field utilities
+
+# A bit of type piracy to allow `Field`s to be indexed with tuples
+# TODO: Consider proposing this as a change in Oceananigans
+@inline @propagate_inbounds Base.getindex(field::Field, idx::NTuple{2, Integer}) = field[idx...]
+@inline @propagate_inbounds Base.getindex(field::Field, idx::NTuple{3, Integer}) = field[idx...]
```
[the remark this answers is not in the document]
bgroenks96: Yes, so this it he point where we have to decide, I think, whether to lean into the more generic indexing syntax or revert back to just using `i, j, k` everywhere. This hacky dipsatch is meant to temporarily facilitate the former, but should ideally be in Oceananigans upstream if at all.

src/processes/surface_energy/albedo.jl
maximilian-gelbrecht: Ok. 

Btw: in a coupled land-atmosphere model would snow be still fully in the land model? Or how would that work?

bgroenks96: Yes snow is fully in the land model. Sometimes it's implemented as a separate tile, but I would prefer to treat it as a dynamic property of the land surface that can occur on (most) tiles, with the exception of ocean shelves. Initially we'll implement a simple bulk snow scheme that treats the entire snow pack as a single layer with homogeneous properties and a corresponding coverage fraction for the grid cell.

From there we can consider two routes: ML-based parameterization of snow processes or a more comprehensive physical treatment of snow that vertically discretizes the snowpack and treats it as a moving boundary problem, possibly utilizing the z-star coordinates (similar to Speedy's sigma coordinates) built into Oceananigans.

src/processes/abstract_types.jl
bgroenks96: Fixed

.github/workflows/examples_CI.yml
```diff
@@ -12,7 +12,7 @@ jobs:
     strategy:
       matrix:
         version:
-          - '1' # test always on the latest released version 
+          - '1.10'
```
maximilian-gelbrecht: because of Oceananigans?

src/abstract_variables.jl
```diff
@@ -141,6 +141,9 @@ struct InputVariable{VD<:VarDims, UT<:Units} <: AbstractVariable{VD}
     "Physical untis associated with this state variable"
     units::UT
 
+    "Interval domain on which scalar fields of this variable are defined"
+    domain::DT
```
maximilian-gelbrecht: What's the domain used for?

src/models/soil/soil_model_bcs.jl
```diff
@@ -50,12 +65,27 @@ ImpermeableBoundary() = NoFlux(:saturation_water_ice)
...
+FreeDrainage(::Type{NF}) where {NF} = PrescribedGradient(:pressure_head, zero(NF))
+
+# SoilBoundaryConditions constructor
 
 """
-Alias for `ColumnBoundaryConditions` with defaults suitable for `SoilModel`s.
+Creates `ColumnBoundaryConditions` with defaults suitable for the given soil process configurations.
 """
-SoilBoundaryConditions(::Type{NF}; top=default_soil_upperbc(NF), bottom=default_soil_lowerbc(NF)) where {NF} = ColumnBoundaryConditions(; top, bottom)
+SoilBoundaryConditions(
```
maximilian-gelbrecht: I find it a bit confusing that there is `SoilBC` and `SoilBoundaryConditions`. The former is just one boundary, correct? And `SoilBoundaryConditions` just wraps around it to initialise top and bottom boundary conditions for one column.  

What's actually also our convention of abbreviations in types? I am rather in favour of writing out abbreviations in types. `SoilBoundaryCondition` and `SoilBoundaryConditions` are obviously also very easy to mistake for each other though.

src/processes/abstract_types.jl
```diff
@@ -14,14 +14,129 @@ variables(op::AbstractOperator) = ()
...
+
+abstract type AbstractSoilEnergyBalance{NF} <: AbstractProcess end
+
+abstract type AbstractSoilHydrology{NF} <: AbstractProcess end
+
+abstract type AbstractSoilBiogeochemistry{NF} <: AbstractProcess end
+
+# Vegetation process types
+
+abstract type AbstractPhotosynthesis end
```
maximilian-gelbrecht: Unrelated to this PR, but why are they not `<:AbstractProcess`?

src/processes/physical_constants.jl
```diff
@@ -20,7 +26,34 @@ A collection of general physical constants that do not (usually) need to be vari
...
 end
 
 PhysicalConstants(::Type{NF}; kwargs...) where {NF} = PhysicalConstants{NF}(; kwargs...)
+
+"""
+    celsius_to_kelvin(c::PhysicalConstants, T)
+
+Convert the given temperature in °C to Kelvin based on the constant `Tref`.
+"""
+celsius_to_kelvin(c::PhysicalConstants, T) = T + c.Tref
```
maximilian-gelbrecht: `@inline` probably

src/processes/prescribed_atmosphere.jl
maximilian-gelbrecht: Looks good to me!

src/abstract_variables.jl
bgroenks96: I think I added it to `InputVariable` just so the constructors would be consistent. Omitting it made some logical sense (don't really need it for inputs) but caused some issues elsewhere, though I can't remember exactly what it was.

src/state_variables.jl
```diff
@@ -72,12 +98,15 @@ function StateVariables(
     )
 end
 
-# TODO: just take the first prognostic variable for the eltype, bad idea?
-Base.eltype(vars::StateVariables) = eltype(vars.prognostic[1])
```
bgroenks96: @maximilian-gelbrecht Turns out this was in fact a bad idea. There can be model configurations (especially in tests) that do not have prognostic variables. Sorry, should have caught that before.

But I went ahead and re-added the `NF` type argument to `StateVariables`.

PR #53, opened 2025-11-19
Pluto notebook in docs

maximilian-gelbrecht:
I very much like the idea of having Pluto notebooks as examples and in the docs. 

It makes the docs more complete when they are there, but at the same time they can also serve as actual notebooks when people want to play around with the code. 

This is just a first draft, let's see if this already works. Workflows are a bit hard to test locally.


maximilian-gelbrecht:
Finally it works. That took too long.

bgroenks96 (approved):

PR #54, opened 2025-11-20
Initial surface hydrology implementation

mahabadri:
This PR provides an initial implementation of the surface hydrology, including canopy water and canopy rain interception from PALADYN

### DIFF
--- a/src/processes/surface_hydrology/canopy_hydrology.jl
+++ b/src/processes/surface_hydrology/canopy_hydrology.jl
@@ -8,7 +8,7 @@ $TYPEDFIELDS
 """
 @kwdef struct PALADYNCanopyHydrology{NF} <: AbstractCanopyHydrology
     
-    "Canopy water interception factor"
+    "Canopy water interception factor for tree PFTs"
     α_int::NF = 0.2 
 
     "Extinction coefficient for radiation"
@@ -27,10 +27,10 @@ variables(::PALADYNCanopyHydrology) = (
     prognostic(:w_can, XY(); desc="Canopy liquid water", units=u"kg/m^2"), 
     auxiliary(:I_can, XY(); desc="Canopy rain interception", units=u"kg/m^2/s"), 
     auxiliary(:E_can, XY(); desc="Canopy evaporation", units=u"kg/m^2/s"),
+    auxiliary(:precip_ground, XY(); desc="Rainfall rate reaching the ground", units=u"kg/m^2/s"),
     input(:LAI, XY(); desc="Leaf Area Index", units=u"m^2/m^2"), 
     input(:SAI, XY(); desc="Stem Area Index", units=u"m^2/m^2")
 )
-
 
 """
     $SIGNATURES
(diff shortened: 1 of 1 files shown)


src/processes/surface_hydrology/canopy_hydrology.jl
```diff
@@ -0,0 +1,98 @@
...
+    return I_can
+end
+
+"""
+    $SIGNATURES
+Computes the `w_can` tendency following Eq. 41, PALADYN (Willeit 2016).
+Modified to ignore the evaporation term for now.
+"""
+
+@inline function compute_w_can_tend(canopy_hydrology::PALADYNCanopyHydrology{NF}, Precip_ground, LAI, SAI, w_can) where NF
```
maximilian-gelbrecht: style-wise I'd avoid capitalisation for variables like `precip_ground` or `precip` below. The precipitation from `PrescribedAtmosphere` is called `rainfall` so far btw.

src/processes/surface_hydrology/canopy_hydrology.jl
```diff
@@ -0,0 +1,98 @@
...
+    "Canopy water removal timescale"
+    τ_w::NF = 86400.0 # [s]
+end
+
+PALADYNCanopyHydrology(::Type{NF}; kwargs...) where {NF} = PALADYNCanopyHydrology{NF}(; kwargs...)
+
+variables(::PALADYNCanopyHydrology) = (
+    prognostic(:w_can, XY()), # Canopy liquid water [kg/m²]
+    auxiliary(:I_can, XY()), # Canopy rain interception [kg/m²/s]
+    input(:Precip, XY()), # Precipitation rate [kg/m²/s]
```
bgroenks96: this is redundant with the `AbstractAtmosphere`. You can remove and then use `precipitation(i, j, state, atmos)`. Or at least I think so... unless I forgot to add a method for it.

In general, the rule I think we will try to follow here is this:

- In the case that a process has a "hard" dependency on another process, i.e. we call `get_` whatever and pass the other process type into the kernel, we do not need to replicate their state variables. Rather, we should use methods like the one above to access the relevant variables or parameters.
- If the other process is only a "soft" dependency where there are shared variables but the process type is not needed/used, then the variable should be declared as an input, as is the case with `LAI` below.

I don't think we are following this rule consistently at the moment, but I think it's what we should try to converge on as a general pattern. Also we will need to remember to put this in the docs.

src/models/surface/surface_hydrology_model.jl
```diff
@@ -0,0 +1,81 @@
...
+    GridType<:AbstractLandGrid{NF},
+    Atmosphere<:AbstractAtmosphere,
+    Initializer<:AbstractInitializer,
+} <: AbstractSurfaceHydrologyModel{NF, GridType}
+    "Canopy hydrology scheme"
+    
+    canopy_hydrology::CanopyHydrology = CanopyHydrology()
+    
+    "Runoff scheme"
+    runoff::Runoff = Runoff()
```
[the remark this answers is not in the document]
mahabadri: Here, I just mean surface runoff. I still need to check whether the same approach is usually used for both surface and subsurface runoff, or if different approaches are used.

PR #56, opened 2025-11-25
Checkpointing extension and example

### LINKED ISSUE #10: Improve differentiability test

maximilian-gelbrecht:
This sets up an extension for checkpointing that defines a checkpointed `run!` and a basic example using it. 

Currently the `main` of has a missing compat for DatasStructures, I've run this on my fork of Checkpointing and already made a PR there. 

The output of the sensitivity is correctly centered around the seed point, but very small in scale ($$10^{-7}$$)

I'll wait for some comments from @bgroenks96, but I am thinking in the end, I'll turn it into another Pluto notebook once we are settled on the concrete example and setting. 

Resolves #10


maximilian-gelbrecht:
The CI is also failing due to the missing compat, but that hopefully won't be an issue for long. We could also relax our own compat of `DataStructures` but Checkpointing should also just be updated.

ext/TerrariumCheckpointingExt/TerrariumCheckpointingExt.jl
```diff
@@ -0,0 +1,33 @@
...
+"""
+function Terrarium.run!(
+        integrator::ModelIntegrator,
+        checkpointing_scheme::Checkpointing.Scheme,
+        steps::Int,
+        Δt = default_dt(timestepper(integrator))
+    )
+    Δt = Terrarium.convert_dt(Δt)
+
+    @ad_checkpoint checkpointing_scheme for _ in 1:steps
```
bgroenks96: Could you briefly explain what this actually does?

bgroenks96 (commented):
I didn't quite understand, what is the issue with `DataStructures`?

ext/TerrariumCheckpointingExt/TerrariumCheckpointingExt.jl
maximilian-gelbrecht: The macro applies a checkpointing scheme to the loop it is marking. So intermediate values in the (forward pass of the) AD are only saved at those points. How often a checkpoint is saved depends on the scheme and the setting, both is set in the script. Here I use the `Revolve` scheme (https://dl.acm.org/doi/10.1145/347837.347846) and save the checkpoint at every time step. In this Julia implementation the inner body has to be differentiated by Enzyme, but for the outer body even other AD systems would work.

maximilian-gelbrecht:
I'll turn this now in a Pluto notebook as well with #53 merged.

bgroenks96 (approved):
Feel free to merge after fixing the `Project.toml` conflict.

ext/TerrariumCheckpointingExt/TerrariumCheckpointingExt.jl
maximilian-gelbrecht: There's a really great side-effect of Checkpointing for Enzyme: it massively reduces compile time. 

Without Checkpointing Enzyme would look at building and compiling the computational graph of the complete iteration over all time steps. With Checkpointing Enzyme just differentiates the body of the time loop. It just has to compile this gradient once, and can then recompute intermediate values very efficiently.

bgroenks96 (changes requested):

bgroenks96 (approved):

Issue #66, opened 2026-01-24
Issues with bloated model state and long compile times


bgroenks96:
We currently have a problem with our `StateVariables` type becoming too complex in larger models. As an example, for the current prototype `VegetationSoilModel`, the following code to initialize on CPU
```julia
vegsoil = VegetationSoilModel(grid; soil)
integrator = @time initialize(vegsoil, ForwardEuler());
```
takes more than two minutes to compile:
```
139.127808 seconds (24.78 M allocations: 1.712 GiB, 0.39% gc time, 103.23% compilation time)
```
with the first `timestep!` being similarly bad:
```julia
@time timestep!(integrator, 60.0)
```
```
103.142561 seconds (8.44 M allocations: 672.929 MiB, 0.19% gc time, 113.19% compilation time)
```

Obviously this is a major usability concern, and I don't even want to know how long it will take to compile with Enzyme... I haven't yet dared to try it.

Note the compilation time >100% which likely indicates recompilation is occurring.

Furthermore, benchmarking `adapt` for the state shows a similar problem

```julia
@btime adapt(CUDA.KernelAdaptor(), $integrator.state);
```
```
108.343 μs (636 allocations: 631.53 KiB)
```
which basically means we're losing 108 microseconds on each kernel launch. I haven't actually counted, but we probably have something like around 10-15 kernels being launched in this model configuration on each timestep, which would amount to >1 ms being spent just on adapting the state to GPU...

I would say this is a fairly serious problem that should be addressed ASAP, likely also before releasing a v0.1.

The solution in the end will likely be to abandon our current pattern of passing `state` around into every kernel and instead passing only the `Field`s necessary for the kernel computation. Note that this also how Oceananigans does it. @glwagner and @simone-silvestri 

There are a couple of possible solutions we could look at here that would still preserve some semblance of modularity.

1. Keep the current state pattern but use only a restricted subset of `state`. Following the convention of #65, something like
```julia
newstate = fields(state, process, args...)
launch!(grid, XYZ, compute_auxiliary_kernel!, newstate, args...)
```
where `fields` would need to be implemented for each process type. Ideally it should be able to be implemented automatically using variable info from `variables`; however, I don't think we can do this in a type stable way because the var info is not available at compile time. I think we would need to change `variables` to be defined on `Type`s rather than the concrete objects.
2. Modify kernels to accept named tuples of a restricted subset of fields.
```julia
[...]

PR #68, opened 2026-02-02
Major refactoring of model and process interface

### LINKED ISSUE #65: Proposed changes to model and process interface
## Preliminaries

The existing process and model interface is largely centered around three primary methods: `variables`, `compute_auxiliary!`, and `compute_tendencies!` (also `timestep!` but this is only when paired with a timestepper). Models additionally define `get_grid`, `get_initializer`, `get_closures`, and `processes`.

Both `AbstractProcess`es and `AbstractModel`s define methods `variables(::AbstractModel)` or `variables(::AbstractProcess)` which return a tuple of variable types which fully define their state space; prognostic variables are time-integrated quantities with associated `tendencies` while auxiliary variables are intermediate values that will be used to compute a tendency either in the same process or some other process. Note that auxiliary variables should *not* include purely diagnostic quantities which should rather be implemented using [Diagnostics](https://clima.github.io/OceananigansDocumentation/stable/appendix/library/#Diagnostics) or callbacks to avoid polluting model state.

The current interface for `AbstractProcess`es has each process accept the current `state`, a `model` and the process type itself in `compute_auxiliary!` and `compute_tendencies!`, i.e.

```julia
compute_auxiliary!(state, model, ::ProcessType)
compute_tendencies!(state, model, ::ProcessType)
```

The idea behind this was to keep the process interface clean and compact; the main advantage is that every process is able to be called with the same basic signature.

Models are generally built by composing together one or more processes and/or other models, allowing for some degree of code reuse across different model configurations. Model construction can, to some extent, benefit from the simplicity of the current process interface since the model does not really need to know what other processes are required for each call to `compute_auxiliary!` and `compute_tendencies!`; however, this apparent veil of ignorance is actually quite superficial as will be discussed below.

## Limitations of the current interface

Some (perhaps predictable) problems have become apparent after working with this design for some time:

1. **Obfuscation of dependencies between processes**. Most processes do not operate in a vacuum; they often depend on the parameters or state variables of other processes. While the current design makes these dependencies apparent internally within each process, namely through the arguments of kernels and kernel functions, the simplified externally facing `compute_auxiliary!` and `compute_tendencies!` interfaces obfuscate them to the caller. This makes it difficult to implement coupling logic in more complex models since the order in which processes should be evaluated is implicit. We want to be as explicit as possible in Terrarium!
2. **Process computations are indirectly dependent on models**. In principle, this isn't necessarily a bad thing. The end goal of every physical process is to be part of some larger model, and this pattern has worked relatively well in other packages such as `SpeedyWeather`. However, a key disadvantage of this approach is that it makes individual process implementations harder to unit test. This is because one needs to configure an entire model in order to test a single process, which isn't really ideal. For (unit) testing, we ideally want to (a) be explicit in what exactly we're testing and (b) limit ourselves to only the bare minimum context needed to test the target code. The model-dependent process interfaces violate these guidelines. In addition, it can cause issues with compiler efficiency since changes in the model type trigger recompilation of all process methods.
3. **Ambiguity in coupling logic between processes and submodels**. The current interface leads to some strange coupling logic such as this code in `VegetationSoilModel`:
```julia
compute_auxiliary!(state, model, model.atmosphere)
compute_auxiliary!(state, model.soil)
compute_auxiliary!(state, model, model.surface_energy_balance)
compute_auxiliary!(state, model, model.plant_available_water)
compute_auxiliary!(state, model.vegetation)
compute_auxiliary!(state, model, model.surface_hydrology)
# recompute surface energy fluxes
compute_surface_energy_fluxes!(state, model, model.surface_energy_balance)
```
which, besides obfuscating interdependencies implicit in the call order (see point 1), also exhibits a weird and unintuitive mixing of the model and process interface. Here I was able to get away with just using the `compute_auxiliary!` dispatches for the models outright, but this might not always be the case. Indeed sometimes, under the current system, the top-level model might need to manipulate the call order of the processes contained within the submodels. While that is possible to do here, it is by no means intuitive and to some extent undermines the entire point of having submodels in the first place.
4. **Duplication of processes and data structures in submodels**. Since models are always designed to be runnable standalone (as implied by the `compute_*!(state, model)` interface), they always require at minimum a `grid`, an `initializer`, and usually the `constants::PhysicalConstants` struct. This results in these objects being duplicated across larger coupled model structures; while this is not necessarily a performance problem per-se (Julia always uses copy-by-reference), it makes model construction much less intuitive because the user needs to make sure that they pass the same `grid` or `constants` object to all of the various submodels. It also greatly complicates model destructuring and restructuring which is necessary for parameter handling.

### Model-free method interface for processes

I propose to modify the process interface to (a) be fully decoupled from models, and (b) make interdependencies explicit. The new interface for the `compute_*!` methods would thus be
```julia
compute_auxiliary!(state, grid, proc::AbstractProcess, args...)
compute_tendencies!(state, grid, proc::AbstractProcess, args...)
```
where `args` are zero or more process or parameterization types which `proc` requires for its computations. Note that `args` may differ for `compute_auxiliary!` and `compute_tendencies!`.

In addition to this, the interface will also have an implicit convention: **if process `A` is passed in the `args` to process `B`, then the corresponding `compute_*!` method for process `A` must have already been called.** This simple convention, in combination with the explicit interface, will make clear the necessary call order and thus help to ensure correctness and physical consistency.

Note that one difficulty that may arise here is in the case where process `B` needs parameters from process `A` but process `A` needs auxiliary variable output from process `B`. This would require a violation of the aforementioned convention in order to be resolved. This is potentially acceptable in some cases but such circular dependencies likely indicate a deeper design issue that should be addressed. A better solution would be to split up those processes into smaller components that have a clearer dependency structure.

### Au revoir, submodels, bonjour coupled processes

As described above (point 4), the submodel concept has proven difficult to implement in practice. I propose to abandon submodels so that every model will consist only of processes (and a `grid`, `constants`, etc.). However, to avoid ending up with coupled models that have dozens of fields corresponding to different processes, we can organize processes into midlevel structures that serve as abstractions for the top-level model. We already do this with `SurfaceEnergyBalance` and `SurfaceHydrology` which group together several subprocesses into one process type.

The primary change here would be threefold:
1. We formalize the notion of coupled process types into a new abstract type `AbstractCoupledProcesses <: AbstractProcess` which additionally includes `processes` in the method interface.
2. We apply the same change to the process interface to coupled process types thus making explicit which (if any) external processes they depend on.
3. We add coupled process types for soil and vegetation which would partially replace the existing `SoilModel` and `VegetationModel` with e.g. `SoilEnergyHydrology` and `VegetationCarbon` (names subject to change).

Note that another potential benefit of having coupled process types is that we could potentially use them to implement kernel fusion for specific combinations of processes. I still think that every process should be independently runnable, and thus should define its own kernels; however, at the cost of some minor redundancy, we could also have specialized kernels for coupled processes that bypass the individual process ones and combine multiple calculations into a single kernel wherever possible.

## Pitfalls

The proposed changes are not without trade-offs. One obvious drawback is that the call-interface for processes becomes more complex and less generic; i.e. the caller (usually a model) has to explicitly choose a valid call pattern for each process and provide all of the necessary dependencies. In general, this is a good thing and is more complex by design. However, it does mean that the model code becomes less agnostic to the concrete types/implementations of each of its processes and thus less generic.

We can control this complexity, at least to some extent, by having standardized interfaces for each process hierarchy. As an example, `AbstractSoilHydrology` would have a finite number of defined call patterns implemented by subtypes, each of which constitutes a particular coupling with one or more other processes. In the simplest case, the only other process dependencies would be other soil processes like `AbstractStratigraphy` and `AbstractSoilBiogeochemistry`. But there could (and already are) other dispatches that might take in, e.g. `AbstractEvapotranspiration` which would then add another forcing term for ET. Not all process implementations necessarily have to support all coupling interfaces; e.g. `GroundEvaporation` does not need to support the interface defined by `CanopyEvapotranspiration` and vice versa. In these cases, its up to the calling code to decide which interface should be used.

bgroenks96:
Since @maximilian-gelbrecht wished for it, here's another mega-PR :nerd_face: 

The main objectives of this PR are to address #65 and (partially) #66 via an extensive refactoring the core model/process interface.

## Key changes

1. **Decoupling of Process and Model interfaces**. Previously, process implementations of the core interface methods like `compute_auxiliary!` and `compute_tendencies!` accepted the `model` as an argument and extracted the relevant components from it (e.g. `grid`, `constants`, and all process dependencies). As discussed in #65, there is nothing inherently wrong with this pattern, but it has some unfortunate drawbacks: (i) it makes coupling interfaces more opaque since the logic of querying and dispatching on the relevant model components are "hidden" inside of the process methods, (ii) it increases compile-time complexity by specializing all `Process` methods on `model`, and (iii) it is incompatible with the proposal in #65 to eliminate submodels in favor of a simpler model structure where there is always one top-level model.

In the refactored interface, implementations of `AbstractProcess` must instead implement methods following the general pattern:
```julia
compute_auxiliary!(state, grid, ::Process, args...)
```
where `grid` is the `AbstractLandGrid` for the model and `args` is zero or more `AbstractProcess` or other parameter `struct`s on which the process depends. The obvious downside to this approach is that the method signatures for all processes are no longer standardized, and thus building models is a bit more difficult (i.e. not just copy-pasting `compute_auxiliary!` and `compute_tendencies!` for each process...). However, the idea is that this can be mitigated by having each `AbstractProcess` subtype define (and document) a finite set of internally standardized calling patterns that corresponding to different coupling cases. Ideally, these different cases should be additive, e.g.

`compute_tendencies!(state, grid, ::SoilHydrology, soil::AbstractSoil, constants::PhysicalConstants, evtr::Optional{AbstractEvapotranspiration} = nothing)`

This pattern implies that evapotranspiration is an optional, additional process which can be handled when explicitly supplied by the calling code. Implementations of `SoilHydrology` that do not handle it can define dispatches like:

`compute_tendencies!(state, grid, ::SoilHydrology, soil::AbstractSoil, constants::PhysicalConstants, args...)`

to remain forward-compatible.

2. **No more `state` in kernels**. This is probably the most pervasive part of this refactoring effort. To avoid doing much of the same refactoring work twice (since we're already redefining interfaces), I decided to go ahead and make this part of the same PR. This is a response to the problems outlined in #66 with long compile times and unacceptable kernel adaptation overhead in coupled models. The basic idea here is that state `Field`s need to be explicitly unpacked and passed into kernels rather than just passing in the entire model `state`. This becomes very important for larger models where the `state` may contain quite a few `Field`s. The general pattern can be summarized by the following excerpt from the refactored soil energy implementation:

```julia
function compute_tendencies!(
    state, grid,
    energy::SoilEnergyBalance,
    ground::AbstractGround,
    args...
)
    # Get dependencies
    procs = (get_hydrology(ground), get_stratigraphy(ground), get_biogeochemistry(ground))
    # Get output (tendency) fields
    tendencies = tendency_fields(state, energy)
    # Get other fields (does not include tendencies)
    fields = get_fields(state, energy, procs...)
    launch!(grid, XYZ, compute_energy_tendencies_kernel!, tendencies, fields, energy, procs...)
    return nothing
end

# Mutating outer kernel function
@propagate_inbounds function compute_energy_tendencies!(tendencies, i, j, k, grid, fields, energy::SoilEnergyBalance, args...)
    tendencies.internal_energy[i, j, k] += compute_energy_tendency(i, j, k, grid, fields, energy, args...)
end

# Non-mutating inner kernel function
@propagate_inbounds function compute_energy_tendency(
    i, j, k, grid, fields,
    energy::SoilEnergyBalance{NF, <:ExplicitTwoPhaseHeatConduction},
    hydrology::AbstractSoilHydrology,
    strat::AbstractStratigraphy,
    bgc::AbstractSoilBiogeochemistry
) where {NF}
    # Operators require the underlying Oceananigans grid
    field_grid = get_field_grid(grid)

    # Divergence of heat fluxes
    ∂U∂t = -∂zᵃᵃᶜ(i, j, k, field_grid, diffusive_heat_flux, fields, energy, hydrology, strat, bgc)
    return ∂U∂t
end

@kernel function compute_energy_tendencies_kernel!(tendencies, grid, fields, energy::SoilEnergyBalance, args...)
    i, j, k = @index(Global, NTuple)
    compute_energy_tendencies!(tendencies, i, j, k, grid, fields, energy, args...)
end
```

Note that the separation of the mutating and non-mutating kernel functions is not strictly necessary; one could directly mutate the field in the `@kernel` method or implement everything in the mutating `compute_energy_tendencies!` function. However, the idea here is to buy some code reusability at the cost of some redundancy. I am mostly thinking of the case where we may want to implement fused kernels when multiple processes are coupled together but still allow each individual process to be fully usable standalone. Having minimal kernels and a division between mutating/non-mutating kernel functions helps with this since calling code can "hook in" at whichever abstraction level is appropriate; i.e. top-level `compute_tendencies!`, mutating kernel function `compute_energy_tendencies!` or the nonmutating `compute_energy_tendency`. Happy to get feedback on this though, maybe it's too verbose.

3. **Minimal `@kernel` methods**. This is mostly an extension of point (2), but I think to maximize reusability of code more generally, we should move towards a programming pattern of always keeping `@kernel` methods as minimal as possible and doing all of the actual work in kernel functions. This is also more-or-less in line with the general design philosophy of Oceananigans and NumericalEarth (@glwagner).

## New general functionality

To facilitate the above refactoring, there are several new utilities related to state variable handling:

1. `prognostic_variables(obj)`, `auxiliary_variables(obj)`, etc. These methods collect and deduplicate all variables of the given type defined by the given  `obj` (model or process). This is mostly useful for the following,
2. `get_fields(state, components...)`, `prognostic_fields(state, components...)`, `auxiliary_fields(state, components...)`, etc. These methods allow you to extract all `Field`s for a particular variable group (prognostic, auxiliary, closure, inputs) defined on the given arbitrary set of components. This lets you filter out a smaller subset of model variables that are (potentially) needed for any given kernel operation. Crucially, these methods are fully type stable thanks to some magic with named tuples and the `AbstractVariable` types.
3. Closures now directly define `variables` as first-class model objects and closure variables are treated as an (implicit) and queryable variable group.

### Debug hooks

Additionally, since [bb53a0a](https://github.com/TUM-PIK-ESM/Terrarium.jl/pull/68/commits/bb53a0a2143fe0808af8c8354556a878a76bb504) there are now automated debugging hooks that should help a lot with tracking down numerical issues. This includes four new diagnostic methods:
- `nancheck!(field_or_nt)` for `Field`s and `NamedTuple`s of `Field`s which forwards to `Oceananigans.Diagnostics.hasnan` and raises an error if `NaN`s are detected.
- `debug!(::Bool)` enables or disables debug-mode by setting a global variable `Terrarium.DEBUG`.
- `debugsite!(args...)` is called from within model code (generally from outside of a kernel) and checks the global variable `Terrarium.DEBUG`; if `false`, the method does nothing and immediately returns. Otherwise, it forwards the call to `debughook!`. Note that invoking `debugsite!` when debug mode is disabled has virtually zero runtime cost and is type stable.
- `debughook!` is the receiver for `debugsite!` calls. Developers can define dispatches for `debughook!` that match corresponding `debugsite!` calls to quickly and easily "hook" into specific parts of the model without having to directly modify the source code. Currently, all kernel `launch!` invocations call `debugsite!` with the kernel method and its arguments, so a corresponding `debughook!` for kernel debugging would then be:
```julia
function Terrarium.debughook!(
    ::typeof(Terrarium.compute_auxiliary_kernel!),
    out, grid, fields,
    proc::ProcessType,
    args...
)
    # Install Infiltrator in global environment;
    # you could also use the debugger but sometimes Infiltrator is faster/easier to use
    Infiltrator.@infiltrate
end
```
Terrarium currently provides default `debughook!` implementations for `compute_auxiliary_kernel!` and `compute_tendencies_kernel!` which call `nancheck!` on the first kernel arg (usually `out`), which is a very nice side effect made possible by this PR. Thus, after calling `Terrarium.debug!(true)` you can run a model or coupled processes and quickly get some information about which kernel is producing `NaN`s in the output.

Note that this is NOT intended to be redundant with the `NaNChecker` callback in Oceananigans (:eyes:  @glwagner). Terrarium does not yet support Oceanangians callbacks but will soon. The `NaNChecker` will still be the preferred "first line of defense" for catching `NaN`s in `Simulation`s with the callback invoked after `compute_tendencies!`, `update_state!`, or `timestep!`. The `debughook!` utility in Terrarium is intended to provide a more granular level of debugging control to hook into specific kernel launches or computations within the model.

## Changes to processes

1. Soil porosity parameters and functions are moved to `Stratigraphy` instead of `SoilHydraulics`. This makes more sense since ultimately it's a property of the soil not a direct consequence of hydrology (though of course in reality the two are closely related).
2. `Stratigraphy` is now "demoted" to a parameterization rather than an `AbstractProcess`. Of course, in the future, we may consider some variables like porosity or soil structure to be prognostic, but these would likely be implemented as separate processes with which `Stratigraphy` is coupled, not as completely separate `Stratigraphy` formulations.
3. The `SurfaceEnergyBalance` is now implemented as a single fused kernel.
5. The `AbstractProcess` type now includes `NF` as a mandatory type parameter (as it always should have been...). All process implementations, including those with no parameters, are updated to follow this pattern.

Resolves #65

## TODO
- [x] Refactor and test soil processes
- [x] Refactor and test vegetation processes
- [x] Refactor and test surface energy balance
- [x] Refactor and test surface hydrology
- [x] Verify differentiability and GPU compatibility

### DIFF
--- a/src/processes/atmosphere/prescribed_atmosphere.jl
+++ b/src/processes/atmosphere/prescribed_atmosphere.jl
@@ -8,8 +8,10 @@ end
 Base.nameof(::TracerGas{name}) where {name} = name
 
 variables(::TracerGas{name}) where {name} = (
-    input(name, XY(), default=380, units=u"ppm", desc="Ambient atmospheric $(name) concentration in ppm"),
+    input(name, XY(), default=default_tracer_conc(Val(name)), units=u"ppm", desc="Ambient atmospheric $(name) concentration in ppm"),
 )
+
+default_tracer_conc(::Val{:CO2}) = 380 # in ppm
 
 """
 Creates a `TracerGas` for ambient CO2 with concentration prescribed by an input variable with


src/abstract_model.jl
```diff
@@ -10,37 +10,46 @@ be able to operate on any given set of `Field`s and parameters representing the
...
+abstract type AbstractProcess{NF} end
+
+"""
+    $TYPEDEF
 
-# AbstractModel interface
+Base type for model components that are responsible for coupling two or more
+`AbstractProcess`es together.
+"""
+abstract type AbstractComponent{NF} <: AbstractProcess{NF} end
```
bgroenks96: @maximilian-gelbrecht and @mahabadri Happy to get feedback on the naming here... `AbstractComponent` might be a bit vague. I also considered `AbstractCoupledProcesses` and `AbstractSystem`.

src/abstract_model.jl
```diff
@@ -64,145 +73,90 @@ function initialize! end
...
 """
-    processes(::AbstractModel)
+    $TYPEDSIGNATURES
 
-Return a tuple of `AbstractProcess`es defind by the given `model`.
+Return a tuple of `AbstractProces`es contained in the given model or coupled processes type.
+Note that this is a type-stable, `@generated` function that is compiled for each argument type.
 """
-processes(::AbstractModel) = ()
+@generated function processes(obj::Union{AbstractComponent, AbstractModel})
```
bgroenks96: Note that `processes` now has a default automatic implementation using `@generated`.

src/abstract_variables.jl
```diff
@@ -277,6 +275,18 @@ function Variables(vars::Tuple{Vararg{Union{AbstractProcessVariable, Namespace}}
     )
 end
 
+"""
+    $TYPEDSIGNATURES
+
+Removes duplicate variables from the given tuple of `vars` by name.
+"""
+@generated function deduplicate_vars(vars::Tuple{Vararg{<:AbstractVariable}})
```
bgroenks96: TODO: make more clear in the docstring that this is the type-stable variant of `deduplicate` specifically for `AbstractVariable`s.

src/processes/soil/energy/soil_energy_closures.jl
```diff
@@ -0,0 +1,134 @@
...
+    # have to assume here that the liquid water fraction is zero if T < 0 and one otherwise. This method
+    # should therefore only be used for initialization and should **not** be involved in the calculation
+    # of tendencies.
+    liq = out.liquid_water_fraction[i, j, k] = ifelse(
+        T >= zero(T),
+        one(sat),
+        zero(sat),
+    )
+    # add liquid water fraction to fields
+    fields = merge(fields, (; liquid_water_fraction = out.liquid_water_fraction))
```
bgroenks96: This is an annoying consequence of being more rigorous with partitioning variables.

src/processes/soil/soil_coupled.jl
```diff
@@ -0,0 +1,86 @@
+struct SoilEnergyHydrologyBGC{
```
bgroenks96: @maximilian-gelbrecht and @mahabadri Do you have any better ideas for naming conventions of these coupled process types? I decided here to just be explicit and state what processes are included, but that is admittedly a bit verbose. I also considered `SoilColumn`, `SoilProcesses`, or just `Soil`, but these all seemed a bit vague and not specific. Similarly for vegetation I figured we would have `VegetationCarbon` for the carbon-only processes...? Or maybe just `NaturalVegetation`?

src/processes/soil/energy/soil_energy.jl
```diff
@@ -68,225 +56,124 @@ function ground_temperature(energy::SoilEnergyBalance, grid, clock, fields)
...
+    args...
+)
+    tendencies.internal_energy[i, j, k] += compute_energy_tendency(i, j, k, grid, fields, energy, args...)
+end
+
+@inline function compute_energy_tendency(
+    i, j, k, grid, fields,
     energy::SoilEnergyBalance{NF, <:ExplicitTwoPhaseHeatConduction},
     hydrology::AbstractSoilHydrology,
     strat::AbstractStratigraphy,
```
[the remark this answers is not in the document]
maximilian-gelbrecht: I like being explicit here.

src/processes/soil/energy/soil_energy_closures.jl
maximilian-gelbrecht: See the other comment. Can you explain a bit more the logic behind that?

examples/scripts/vegetation_soil_column.jl
```diff
@@ -5,7 +5,7 @@ using CUDA
 grid = ColumnGrid(GPU(), ExponentialSpacing(Δz_max=1.0, N=30))
 # Set up Richards model for soil hydrology
 swrc = VanGenuchten(α=2.0, n=2.0)
-hydraulic_properties = ConstantHydraulics(eltype(grid), cond_unsat=UnsatKVanGenuchten(eltype(grid); swrc))
+hydraulic_properties = ConstantSoilHydraulics(eltype(grid), cond_unsat=UnsatKVanGenuchten(eltype(grid); swrc))
```
glwagner: out of curiosity, what does "cond unsat" mean?

src/processes/atmosphere/prescribed_atmosphere.jl
```diff
@@ -203,18 +203,18 @@ variables(::LongShortWaveRadiation) = (
 )
 
 """
-    shortwave_in(i, j, grid, state, ::AbstractAtmosphere{PR, <:LongShortWaveRadiation})
+    shortwave_down(i, j, grid, state, ::AbstractAtmosphere{PR, <:LongShortWaveRadiation})
 
 Retrieve or compute the incoming/downwelling shortwave radiation at the current time step.
 """
-shortwave_in(i, j, grid, state, ::AbstractAtmosphere{PR, <:LongShortWaveRadiation}) where {PR} = state.surface_shortwave_down[i, j]
+shortwave_down(i, j, grid, state, ::AbstractAtmosphere{PR, <:LongShortWaveRadiation}) where {PR} = state.surface_shortwave_down[i, j]
```
glwagner: note that in NumericalEarth we use the term "downwelling_shortwave" and "downwelling_longwave"

src/processes/soil/soil_coupled.jl
mahabadri: I think I like it the way it is. At most, it can only include all three “cycles,” so the maximum would be `SoilEnergyHydrologyCarbon`. For vegetation processes, they’re actually only the carbon processes, because we decided that canopy hydrology is part of surface hydrology and canopy energy will be part of surface energy. Also, there are a lot of processes shared between natural and managed vegetation, so I would keep it as “vegetation.”

examples/scripts/vegetation_soil_column.jl
bgroenks96: Changed to `unsat_hydraulic_cond` in [e229d50](https://github.com/TUM-PIK-ESM/Terrarium.jl/pull/68/commits/e229d506830dc303151f92b618e58a6f1a35c54a). Is that better or do you think we should really go for the fully spelled out form? @glwagner

src/models/surface/surface_energy_model.jl
```diff
@@ -31,14 +40,11 @@ function SurfaceEnergyModel(
     return SurfaceEnergyModel(grid, atmosphere, surface_energy_balance, constants, initializer)
 end
 
-processes(model::SurfaceEnergyModel) = (model.atmosphere, model.surface_energy_balance)
```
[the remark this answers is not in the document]
maximilian-gelbrecht: In the same spirit: what's now our convention with getter functions and `processes`? I've seen you removed most getter functions.

src/processes/soil/energy/soil_energy.jl
```diff
@@ -9,54 +7,44 @@ Represents an explicit formulation of the two-phase heat conduction operator in
...
-    ::SoilEnergyBalance{NF, OP, FreeWater},
-    ::AbstractSoilHydrology
-) where {NF, OP} = FreeWater()
-
-get_closure(energy::SoilEnergyBalance) = energy.operator.closure
+) where {NF} = SoilEnergyBalance(operator, closure, thermal_properties)
 
 variables(energy::SoilEnergyBalance) = (
-    prognostic(:internal_energy, XYZ(); closure=get_closure(energy), units=u"J/m^3", desc="Internal energy of the soil volume, including both latent and sensible components"),
-    auxiliary(:liquid_water_fraction, XYZ(), domain=UnitInterval(), desc="Fraction of unfrozen water in the pore space"),
```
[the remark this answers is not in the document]
maximilian-gelbrecht: I see it's defined for below for the closure type, but why does `SoilEnergyBalance` not do a `variables(closure)...` then?

src/processes/soil/energy/soil_energy_closures.jl
```diff
@@ -0,0 +1,134 @@
...
+    # calculate temperature from internal energy and liquid water fraction
+    T = out.temperature[i, j, k] = energy_to_temperature(fc, U, Lθ, C)
+    return T
+end
+
+"""
+Calculate the unfrozen water content from the given internal energy, latent heat content, and saturation.
+"""
+@inline function liquid_water_fraction(::FreeWater, U::NF, Lθ::NF, sat::NF) where {NF}
+    return if U >= zero(U)
```
maximilian-gelbrecht: Unrelated to this change, but I remember that in our initial version we had a problem with Enzyme and some if/else clauses like this. I don't remember how we resolved it. Do all of these ended up working?

src/state_variables.jl
```diff
@@ -112,13 +118,18 @@ function reset_tendencies!(state::StateVariables)
...
 
 Retrieves fields with names given in `queries` and returns them in a `NamedTuple`. Each argument
 in `queries` can either be a `Symbol` corresponding to a field/variable defined in the namespace
 of `state` or a `Pair{Symbol, Tuple}` where the key is the child namespace and the value is a
 tuple of queries from that namespace.
 
+!!! warning
+    This method relies on runtime dispatch and thus should not be used in performance-critical code.
+    If you need to query fields for specific sets of variables or components, use one of the
+    type-stable variants instead.
```
maximilian-gelbrecht: type-stable are literally all other ones?

src/processes/soil/hydrology/soil_hydraulic_closures.jl
```diff
@@ -0,0 +1,93 @@
...
+"""
+    $TYPEDEF
+
+Represents a closure relating saturation of water/ice in soil pores to a corresponding
+pressure (or hydraulic) head. Note that here "pressure head" is defined to be synonymous
+with hydraulic head, i.e. including all both elevation and hydrostatic pressure contributions.
+This relation is typically described by soil property-dependent *soil-water retention curve* (SWRC)
+which is here defined in [`SoilHydraulics`](@ref).
+"""
+@kwdef struct SaturationPressureClosure <: AbstractSoilWaterClosure end
```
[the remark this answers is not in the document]
bgroenks96: A closure is a parameterization type that defines a constitutive relationship between a prognostic variable and an auxiliary variable. It is not an `AbstractProcess` because it's really just a function that also may define some associated variables.

I agree that the current change is confusing, I'll circle back to this once I get the rest of the tests working.

Also worth noting that I think the line between process and parameterization types is uncomfortably blurry at the moment... happy to hear suggestions on how we could improve that abstraction.

src/state_variables.jl
```diff
@@ -128,20 +139,115 @@ fields = get_fields(state, :temperature, :saturation_water_ice)
...
+Retrieves all `Field`s from `state` corresponding to auxiliary variables defined on the given `components`.
+"""
+@inline function auxiliary_fields(state, components...)
+    return get_fields(state, mapreduce(auxiliary_variables, tuplejoin, components))
+end
+
+"""
+    $TYPEDSIGNATURES
+
+Retrieves all `Field`s from `state` corresponding to closure variables defined on the given `components`.
```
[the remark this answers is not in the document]
bgroenks96: Yes, the subset of auxiliary variables that are defined by the closures.

I was also wondering whether or not we should just promote closure variables to be their own distinct group because I do agree that this overlap is confusing.

src/processes/soil/energy/soil_energy_closures.jl
maximilian-gelbrecht: Huh.. okay

maximilian-gelbrecht (commented):
Still not fully through everything, will continue Monday.

maximilian-gelbrecht:
> > Did you already apply a formatter like runic btw? Now would be a good time with this PR.
> 
> I was actually hoping you would do that in #49 after this is merged? 😅



> > Did you already apply a formatter like runic btw? Now would be a good time with this PR.
> 
> I was actually hoping you would do that in #49 after this is merged? 😅

OK sure, I can do it.

bgroenks96:
Here are the benchmarks for state adaptation:

```julia
# Benchmark adapt for full state
res1 = @benchmark adapt(CUDA.KernelAdaptor(), $integrator.state)
```
```
BenchmarkTools.Trial: 10000 samples with 1 evaluation per sample.
 Range (min … max):  103.476 μs …   3.817 ms  ┊ GC (min … max): 0.00% … 92.79%
 Time  (median):     113.430 μs               ┊ GC (median):    0.00%
 Time  (mean ± σ):   127.607 μs ± 146.873 μs  ┊ GC (mean ± σ):  9.43% ±  7.79%

                ▁▄▇██▆▃▁                                         
  ▂▁▁▂▂▂▂▂▂▂▂▃▄▆█████████▇▆▆▅▄▄▄▃▃▃▃▃▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▁▂▂▂▂▂ ▃
  103 μs           Histogram: frequency by time          135 μs <

 Memory estimate: 635.91 KiB, allocs estimate: 634.
```
Now we simulate `adapt` for a kernel launch for soil hydrology that extracts only the relevant fields with `get_fields`:
```julia
# Extract fields for soil hydrology
fields = get_fields(integrator.state, hydrology, soil.biogeochem)
tendencies = Terrarium.tendency_fields(integrator.state, hydrology)
res2 = @benchmark adapt(CUDA.KernelAdaptor(), ($tendencies, $fields))
```
```
BenchmarkTools.Trial: 10000 samples with 9 evaluations per sample.
 Range (min … max):  2.705 μs … 774.532 μs  ┊ GC (min … max): 0.00% … 98.84%
 Time  (median):     3.965 μs               ┊ GC (median):    0.00%
 Time  (mean ± σ):   4.057 μs ±  16.895 μs  ┊ GC (mean ± σ):  9.24% ±  2.21%

     ▅█▇▄                                                      
  ▃▅█████▇▄▃▂▂▂▂▂▂▂▂▂▂▂▂▁▂▂▂▂▂▂▂▃▂▃▃▄▄▄▅▅▆▆▆▆▇▇▆▆▆▆▅▅▄▄▄▄▃▃▃▂ ▃
  2.71 μs         Histogram: frequency by time        4.91 μs <

 Memory estimate: 4.42 KiB, allocs estimate: 51.
```
so I'd say that's a pretty decent improvement :)

src/diagnostics/debugging.jl
maximilian-gelbrecht: I like the idea of some utilities to ease debugging. 

That being said, I haven't tried this yet, but I do wonder a little bit what's the advantage compared to just putting Infiltrator directly somewhere.

examples/notebooks/example_model_notebook.jl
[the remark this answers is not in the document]
bgroenks96: Yes. Should we do that in this PR or in the docs PR?

bgroenks96:
@maximilian-gelbrecht [a3211de](https://github.com/TUM-PIK-ESM/Terrarium.jl/pull/68/commits/a3211de8d02be0d11c449fd683066cb3d3c6bfdf) refactors all of the vegetation processes to use fully spelled out variable names for the state variables instead of abbreviations, as we previously discussed.

src/processes/soil/energy/soil_energy.jl
```diff
@@ -68,225 +56,124 @@ function ground_temperature(energy::SoilEnergyBalance, grid, clock, fields)
...
+    # Get output (tendency) fields
+    tendencies = tendency_fields(state, energy)
+    # Get other fields (does not include tendencies)
+    fields = get_fields(state, energy, procs...)
+    launch!(grid, XYZ, compute_energy_tendencies_kernel!, tendencies, fields, energy, procs...)
     return nothing
 end
 
-@kernel function compute_energy_tendency_kernel!(
+function closure!(
```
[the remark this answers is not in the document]
bgroenks96: might need to be refactored again in the future, I guess... fortunately it really only affects two processes so it's not that hard to refactor.

src/processes/vegetation/plant_available_water.jl
```diff
@@ -34,31 +36,62 @@ function soil_moisture_limiting_factor(::FieldCapacityLimitedPAW, grid, clock, f
...
+    compute!(state.soil_moisture_limiting_factor)
 end
 
-@kernel function compute_paw_kernel!(
-    state, grid,
-    ::FieldCapacityLimitedPAW{NF},
-    hydrology::AbstractSoilHydrology,
+# Kernel functions
+
+@propagate_inbounds function compute_paw(
```
maximilian-gelbrecht: very minor, but when the variable is spelled out `plant_available_water` I'd also spell out the function

src/models/surface/surface_energy_model.jl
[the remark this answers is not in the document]
bgroenks96: To be continued in #71

PR #73, opened 2026-02-09
Runic Formatting (and workflow), but now really 

### LINKED ISSUE #48: Code formatting
After some initial discussion with @bgroenks96 we'll start using code formatter for the repo, most likely [Runic](https://github.com/fredrikekre/Runic.jl/tree/master/.github). I'll make an initial draft PR probably still this week. 

What does this mean? 

What's a code formatter: A code formatter automatically formats code in accordance with a standard style guideline. It's usually a command line tool, but even this can be automated (more below). 

To the existing repo:
* All code will be once formatted in a draft PR 
* A workflow will be added that checks new PRs if they are matching the style, and if not recommend the contributor to format the code (this is not automated
* A small section about code format in the docs 
* (We could add another workflow with automated suggested changes based on the style, but I am not sure we really want this)

For developers: 
* You can add `Runic` to VSCode as a plugin so that either you format on pressing a button or always when saving a file 
* You can add `Runic` as a git hook so that everytime you commit a file to Terrarium it gets formatted automatically 

Both of these things have to be done on the client side, and the readme on https://github.com/fredrikekre/Runic.jl/ provides enough info for that. 

Any questions, concerns?

maximilian-gelbrecht:
Closes #48

### DIFF
--- a/src/models/coupled/vegetation_soil_model.jl
+++ b/src/models/coupled/vegetation_soil_model.jl
@@ -70,7 +70,8 @@ function initialize!(state, model::VegetationSoilModel)
     initialize!(state, grid, model.surface_hydrology)
     # TODO: change when refactoring model/process types
     initialize!(state, grid, model.vegetation, model.atmosphere, model.constants)
-    return initialize!(state, grid, model.soil, model.constants)
+    initialize!(state, grid, model.soil, model.constants)
+    return nothing
 end
 
 function compute_auxiliary!(state, model::VegetationSoilModel)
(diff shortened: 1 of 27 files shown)


examples/scripts/speedy_dry_land.jl
```diff
@@ -16,38 +16,38 @@ Speedy.SpeedyTransforms.FFTW.set_num_threads(1)
...
-    progn::Speedy.PrognosticVariables{NF},
-    diagn::Speedy.DiagnosticVariables{NF},
-    land::TerrariumDryLand,
-    model::Speedy.PrimitiveEquation,
-) where {NF}
+        progn::Speedy.PrognosticVariables{NF},
+        diagn::Speedy.DiagnosticVariables{NF},
+        land::TerrariumDryLand,
+        model::Speedy.PrimitiveEquation,
+    ) where {NF}
```
bgroenks96: I... don't like this. It makes no logical sense. But I guess I just have to get over it...?

src/models/coupled/vegetation_soil_model.jl
```diff
@@ -70,7 +70,7 @@ function initialize!(state, model::VegetationSoilModel)
     initialize!(state, grid, model.surface_hydrology)
     # TODO: change when refactoring model/process types
     initialize!(state, grid, model.vegetation, model.atmosphere, model.constants)
-    initialize!(state, grid, model.soil, model.constants)
+    return initialize!(state, grid, model.soil, model.constants)
```
bgroenks96: I guess we should update all of these methods to `return nothing`.

bgroenks96:
So a few things:

1. I don't understand why the workflow is failing. Maybe you can figure it out?
2. I have learned that I am apparently more opinionated about formatting than I thought I was. Or at least about indentation. I *really* don't like how the multi-line function and type signatures are handled. I saw the rule in Runic's README about using 4-spaces for indentation, which I'm fine with, but I didn't see this... I'll think about whether or not I can get over it.
3. I don't love the `return` thing. I saw this in the README and was skeptical. I agree with it in principle, but I think all the functions that weren't explicitly returning something should just `return nothing` because that's basically what they mean. I think Runic would be compatible with that but we'd have to make the change ourselves. Maybe you can get Claude to do it?

PR #77, opened 2026-02-12
Revise modelling example notebook

maximilian-gelbrecht:
It turned out that the notebook was actually still so much on the surface that nothing really changed. It still runs and the description is also still accurate as far as I can see it. 

But... 

It's probably also good opportunity to change that. We should add an example that actually launches kernels. I was just wondering whether to create a second notebook for that or add it here. Probably just the latter? 

I would just repeat the example but with kernels?


examples/notebooks/example_model_notebook.jl
bgroenks96: This needs to change to `compute_auxiliary!(state, model.grid, model.dynamics)` and same below.

bgroenks96: Same here, this should be `grid`.

bgroenks96: and here

bgroenks96 (changes requested):
This example does define an `AbstractProcess` type whose interface needs to be updated to reflect the new interface.

bgroenks96:
I kind of feel like we should introduce kernels with a slightly more involved example? We don't want to encourage people to use kernels for *everything*. Very simple `Field` operations can and should be done without kernels or with `KernelFunctionOperation` (which we also should introduce somewhere...). Indeed, many `Field` operations have built-in predefined kernels in Oceananigans.

bgroenks96:
> Would we benefit in some way from a common abstract type in our type hierarchy for the parametrizations?

Yeah, I've been considering this for a while. It would be nice from a general interface/type-hierarchy design perspective, but I also have hesitated for exactly the reason that you point out, i.e. there just isn't currently a concrete benefit to having such a common type. The only thing I can think of is that we could enforce the `NF` type parameter, i.e. `AbstractParameterization{NF}`. But other than that, there isn't really any need for a common method interface like there is for processes and models.

bgroenks96:
Oceananigans already has a callback system which is also used in time stepping, e.g. to implement adaptive time stepping. I wouldn't want to implement another callback system on top of that. Can you describe more concretely the problem that you're trying to solve?

maximilian-gelbrecht:
What we talked about yesterday, dispatching on our time stepping to clip negative values in the time step. You can't do that currently as far as I can see it without dispatching also directly on a specific solver, which I don't like at all. 

I think as long as we have our own time steppers and don't use the Oceananigans ones, we might want to have some form of callback system for them.

maximilian-gelbrecht:
I don't know if you've seen it but for Speedy we have a PR to use the artifact system for that now: https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/983

Technically, we might need to add an attribution to ECWMF/ERA5 due to the CC BY licence.

examples/notebooks/example_model_notebook.jl
```diff
@@ -293,34 +367,311 @@ begin
...
+* In this function call we don't hand over the full `StateVariables` anymore as this would come with a large computational overhead on GPUs. Instead we only hand over those fields that the process actually needs. You can either assemble these fields manually, or use the convenience function `get_fields(state, processes...)` that returns all the fields that are defined in the `variables` of the respective processes or the model.
+* Kernels are launched with the `launch!` function that defines whether the kernel is launched over all columns / gridpoints `XY` or over all columns and horizontal layers `XYZ` of a `grid`
+* Kernel functions themselves are supposed to be very minimal functions that forward the actual computation to pointwise compute functions that compute the needed quantity for the grid points `i,j`. This increases the reusability and composability of our model code. These compute functions are really the core of the implementation of our model and similar to those you typically find in more traditional land models: They compute the action of a process for a single grid point given inputs and parameters.
```
bgroenks96:
```suggestion
* Kernel functions themselves are supposed to be very minimal functions that forward the actual computation to pointwise compute functions that compute the needed quantity for the grid points `i,j`. This increases the reusability and composability of our model code. These compute functions are really the core of the implementation of our model and similar to functions you might find in more traditional land model, called from within a loop.They compute the action of a process for a single grid point given inputs and parameters. 
```

PR #78, opened 2026-02-15
Refactor initializer types and make `SoilInitializer` actually useful

bgroenks96:
This PR addresses a longstanding concern that I've had with the current system for specifying initial conditions: namely, that `FieldInitializers` inherently required "inside knowledge" about the model and what `Field`s it would have despite being constructed and provided *before* initialization of the model itself. This is of course a minor issue since the user has to anyway figure this out *eventually*, but I think it does nonetheless constitute a barrier on that initial learning curve. I really started to notice this while writing more detailed documentation explaining the basics of configuring models.

The main change proposed here is to replace `FieldInitializers` with an argument to the `initialize` method, mirroring the current system for `boundary_conditions`; i.e. any `Field`-specific initialization logic would be provided to `initialize` rather than the model itself, which I think makes a lot more sense.

Here is an example of the new API from the global ERA5 script:
```julia
model = SoilModel(grid)
boundary_conditions = PrescribedSurfaceTemperature(:Tair)
# Initial conditions
initializers = (
    # steady-ish state initial condition for temperature
    temperature = (x, z) -> Tsurf_0[Int(round(x)) + 1] - 0.02 * z,
    # dry soil
    saturation_water_ice = 1.0,
)
integrator = initialize(model, ForwardEuler(), Tair_forcing; initializers, boundary_conditions)
```

However, `model`s can and do still define `AbstractInitializer` subtypes that implement model-specific initialization logic and can potentially declare their own parameters and input `Field`s. An example of this is the `SoilInitializer` which was, until now, unused. This PR refactors and improves the documentation of `SoilInitializer` to make it actually useful.

Additional changes:
- Adds pretty printing for model state variables, e.g.
```
julia> variables(model)
Variables
├─ Prognostic: 
├── Variable internal_energy [J m^-3] on XYZ{Center, Center, Center}
├─ Auxiliary: 
├── Variable temperature [°C] on XYZ{Center, Center, Center}
├── Variable liquid_water_fraction [-] on XYZ{Center, Center, Center}
├── Variable ground_temperature [°C] on XY{Center, Center}
├── Variable saturation_water_ice [-] on XYZ{Center, Center, Center}
├── Variable water_table [m] on XY{Center, Center}
├── Variable hydraulic_conductivity [m s^-1] on XYZ{Center, Center, Face}
├─ Inputs: 
├─ Namespaces:
```
- Some other unrelated changes to documentation since this PR grew out of the ongoing documentation effort; I did try to keep it somewhat separated, though.. most of the new documentation changes are in another branch.

### DIFF
--- a/examples/scripts/soil_heat_column.jl
+++ b/examples/scripts/soil_heat_column.jl
@@ -15,7 +15,8 @@ initializer = SoilInitializer(
 model = SoilModel(grid; initializer)
 # constant surface temperature of 1°C
 boundary_conditions = PrescribedSurfaceTemperature(:T_ub, 1.0)
-integrator = initialize(model, ForwardEuler(); boundary_conditions)
+timestepper = ForwardEuler(eltype(grid))
+integrator = initialize(model, timestepper; boundary_conditions)
 # test one timestep
 @time timestep!(integrator)
 # run simulation forward for a set period of time


docs/src/overview/numerical_core.md
```diff
@@ -0,0 +1,75 @@
...
+- [`ColumnRingGrid`](@ref) which is the same as `ColumnGrid` but with columns ordered according to a user-specified `RingGrid` from [RingGrids.jl](https://speedyweather.github.io/SpeedyWeatherDocumentation/stable/ringgrids/). Under this configuration, each dimension along the `X` axis corresponds to a point in the `RingGrid`, allowing for direct translation between Terrarium and RingGrids `Field`s. This is primarily motivated by our goal to couple Terrarium with [SpeedyWeather.jl](https://github.com/SpeedyWeather/SpeedyWeather.jl).
+
+All Terrarium `AbstractLandGrid` implementations are required to implement the following methods:
+- `architecture(grid)` (from Oceananigans) which returns the `Architecture` (e.g. `CPU` or `GPU`) on which the grid is defined,
+- `get_field_grid(grid)` which returns the underlying `Oceananigans` grid.
+
+The vertical `Z`-axis is currently limited to representing the subsurface (typically soil) domain, though we plan to expand this to include snow and canopy layers in the future.
```
maximilian-gelbrecht: Even with what we have at the moment is it actually really limited to soil? Is it not just the only way we currently use the `z` axis? 

Take the `VegetationSoilModel` for example, wouldn't it also just work to define a second `grid_vegetation` as a field and then use it in the calls to the vegetation processes? We might end up doing something a bit more sophisticated like that when really introducing canopy layers, but wouldn't this already work like this?

docs/src/overview/numerical_core.md
```diff
@@ -0,0 +1,75 @@
...
+
+```julia
+@kernel function square_kernel!(output, input, other_args...)
+    i, j, k = @index(Global, NTuple)
+    output[i, j, k] = input[i, j, k]^2
+end
+```
+where `input` and `output` are here assumed to be 3D `Array`s or `Field`s. Note that kernels can also accept any arbitrary number of arguments of any plain data type (including immutable `struct`s) for which `isbits(arg)` is true.
+
+Kernels represent self-contained programs that can be execute in parallel by large number of worker threads or processes. The [`@index`](https://juliagpu.github.io/KernelAbstractions.jl/stable/api/#KernelAbstractions.@index) macro extracts the index of the executing thread within the parallel kernel; `Global` here refers to the index across all workgroups (or "blocks" in CUDA language).
```
maximilian-gelbrecht: I would add something like: 

"In very simple terms you can think of the kernel function as the body of a for loop over all grid points" 

To explain this to people without any strong computational background

docs/src/overview/numerical_core.md
bgroenks96: I guess in theory, yes, but I am still not 100% sure about whether it makes sense to have separate grids for soil and vegetation or just treat them as both living on the same grid with varying spatial discretization. Oceananigans also supports dynamic z-coordinates, like Speedy's sigma-coordinates but even a bit fancier, I think.

I am currently leaning towards the latter because it seems more physically coherent and less error prone, at least if we can come up with a good system for selecting ranges of the grid.

bgroenks96:
@maximilian-gelbrecht is there anything else you'd like added here? or are we ready to merge it?

docs/src/overview/numerical_core.md
maximilian-gelbrecht: I still think that just having a second grid and ensuring they are initialised with the same x,y discretisation would be easier, but let's see when we are at that point. I don't think this will be a major road block in either way.

maximilian-gelbrecht (approved):
No, everything's fine. Feel free to merge

PR #85, opened 2026-03-16
RingGrids input 

maximilian-gelbrecht:
This already adds a `RingGrids.Field` to `Oceananigans.Field` converter. 

For actual usage as inputs, I think am still not super happy with the `InputSource` currently for static input.  

We have `FieldInputSource` that just does nothing for `update_input!` and the user has to copy it over manually on the one hand, and we have `RasterInputSource` that copies the same field at every time step. Both are not optimal. 

Wouldn't it make sense to define a `initialize!` for `InputSources`. This would default to do `nothing`, but static input sources can implement it to load the data once at the start. 

@bgroenks96 what do you think?

### DIFF
--- a/test/grids.jl
+++ b/test/grids.jl
@@ -70,5 +70,29 @@ end
         for k in 1:10
             @test all(ocean_data_3d[:, 1, k] .== ring_field_3d.data[grid.mask.data, k])
         end
+
+        # Test with masked grid (some points inactive)
+
+        # Create a new ring grid for this test
+        mask = rand(Bool, ring_grid)
+        # Create a masked ColumnRingGrid using the correct constructor
+        masked_grid = ColumnRingGrid(UniformSpacing(Δz = 0.5, N = 10), mask)
+
+        ring_field = rand(ring_grid)
+
+        # Convert to Oceananigans field
+        ocean_field_masked = Terrarium.Field(ring_field, masked_grid)
+        @test isa(ocean_field_masked, Terrarium.Field)
+        @test size(ocean_field_masked) == (sum(mask), 1, 1)
+
+        # Verify only masked (active) points were copied
+        ocean_data_masked = Terrarium.interior(ocean_field_masked)
+
+        expected_values = ring_field.data[mask]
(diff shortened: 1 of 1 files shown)


src/grids/column_ring_grid.jl
```diff
@@ -115,6 +115,38 @@ function RingGrids.Field(arch::AbstractArchitecture, field::AbstractVecOrMat, gr
...
+
+    # Create Oceananigans field with appropriate dimensions
+    ndims_extra = ndims(ring_field) - 1  # Extra dimensions beyond the horizontal grid
+    if ndims_extra == 0
+        # 2D field (horizontal only)
+        oceananigans_field = Field{Center, Center, Nothing}(grid.grid)
+        fill!(oceananigans_field, default_value)
+        oceananigans_data = interior(oceananigans_field)
+        oceananigans_data[:, 1, 1] .= ring_field_data[grid.mask.data]
+    else
```
bgroenks96: I think we should simply disallow any `RingGrids.Field` that has more than 2 axes; it's ambiguous what should happen to the extra dimensions and I can only ever imagine it leading to bugs.

bgroenks96:
> We have FieldInputSource that just does nothing for update_input! and the user has to copy it over manually on the one hand, and we have RasterInputSource that copies the same field at every time step. Both are not optimal.

I think the problem here is that `FieldInputSource` wasn't updated after changing the input system. Originally, the `Field`s from the input source were used directly as the `Field`s in `StateVariables`. This was changed when I realized that actually there are legitimate reasons to have separate copies of inputs across namespaces.

Back then, `FieldInputSource` didn't need to do anything when being updated because the `Field` was already used directly. But now, it needs to be initialized.

> Wouldn't it make sense to define a initialize! for InputSources. This would default to do nothing, but static input sources can implement it to load the data once at the start.

I think that's fine, I can't really think of a better solution off the top of my head. I do wonder a bit whether or not models/processes should have control over this, i.e. take `InputSources` in `initialize!` and decide how to set the corresponding input `Field`s. That would lead to a considerable amount of boilerplate, though, and probably wouldn't be needed very often.

PR #103, opened 2026-04-09
 Soil Heat Diffusion Example (#91)

### LINKED ISSUE #91: Create soil heat example based on analytical test case
Add soil heat example as proposed in #88 

Checklist:
- [ ] Add example notebook or script demonstrating how to set up a `SoilModel` to simulate heat conduction for a single column with periodic forcing (see existing [unit test](https://github.com/NumericalEarth/Terrarium.jl/blob/main/test/soil/soil_energy_tests.jl#L89))
- [ ] Show that output matches expected value within some expected tolerance
- [ ] Add documentation page for the example
- [ ] Bonus: Consider also adding a test for the two-phase Stefan problem based on the implementation in [CryoGrid.jl](https://github.com/CryoGrid/CryoGrid.jl/blob/master/test/Solvers/cglite_implicit_tests.jl#L49)

CC @elenamiralles

elenamiralles:
Add soil heat diffusion example with analytical validation

Closes #91


examples/soil_heat_diffusion.jl
```diff
@@ -0,0 +1,122 @@
+# # Soil heat diffusion with periodic surface forcing
+#
+# This example demonstrates how to simulate 1D soil heat conduction
+# in a single vertical column using Terrarium.jl.
+#
+# We verify the numerical solution against the analytical solution
+# of the heat equation with sinusoidal surface forcing:
+#
+#     ∂T/∂t = α ∂²T/∂z²
```
bgroenks96: The forcing isn't included in this equation? Maybe put it on a second line?

examples/soil_heat_diffusion.jl
```diff
@@ -0,0 +1,122 @@
...
+# This example validates:
+#
+# - exponential damping with depth
+# - phase lag with depth
+# - agreement between analytic and numerical solution
+
+using Terrarium
+using CairoMakie
+
+# Physical parameters
```
bgroenks96: Since this will end up getting processed by Literate.jl, you should add Markdown headings in the comments like `# # Physical parameters` or use `##` to make it just a comment.

examples/soil_heat_diffusion.jl
```diff
@@ -0,0 +1,122 @@
...
+    return T
+end
+
+T_sol = heat_conduction_solution(T₀, A, P, α);
+
+# Model setup
+
+grid = ColumnGrid(CPU(), Float64, ExponentialSpacing(Δz_min = 0.05, Δz_max = 100.0, N = 100,));
+
+# Configure soil properties for pure heat diffusion
```
bgroenks96: I think it would be good to write a few sentences before each block to walk the reader through what you're doing.

examples/soil_heat_diffusion.jl
```diff
@@ -0,0 +1,122 @@
...
+bcs = PrescribedSurfaceTemperature(:Tsurf, upper_bc);
+
+# Temp IC
+initializers = (temperature = (x, z) -> T_sol(-z, 0.0),);
+integrator = initialize(model, ForwardEuler(); initializers, boundary_conditions=bcs);
+
+# Run simulation
+
+Δt = 60.0;
+Ts_buffer = [deepcopy(integrator.state.temperature)];
```
bgroenks96: I know this is how it is done in the test, but perhaps here we could use an Oceananigans `Simulation` with `JLD2Writer` and then read the output back in from a file?

bgroenks96 (changes requested):
Looks like a good start, thanks! I left a few comments below.

I am also interested in hearing your feedback; did you run into any issues? Was anything confusing for you or missing clear documentation?

elenamiralles:
Hi @bgroenks96, thank you for the review. I will update the code asap. With respect to documentation everything was very straightforward! One thing that I think could use some improvement is time to download? Maybe it was just my machine but it took 20 min to import the Terrarium package initially.

examples/soil_heat_diffusion.jl
elenamiralles: Attempted this but Oceananigans `Simulation` errors because `SoilModel` doesn't implement `architecture`. Is there a Terrarium specific `Simulation` type planned, or should I hold off on this?

bgroenks96: Can you post the code you used to set it up? You should be passing the `integrator` into `Simulation`, not the `SoilModel`. Sorry, I know that's a bit confusing...

bgroenks96:
> One thing that I think could use some improvement is time to download? Maybe it was just my machine but it took 20 min to import the Terrarium package initially.

That's not good! Do you mean the initial installation of the package? That can definitely be slow in Julia due to precompilation time, but 20 minutes seems excessive to me, unless your internet connection was quite slow.

examples/soil_heat_diffusion.jl
elenamiralles: Ohhh I see. This is what I had previously:

`simulation = Simulation(model; Δt = Δt, stop_time = 2P)`

I've updated the block to:
````
simulation = Simulation(integrator; Δt = Δt, stop_time = 2P)

simulation.output_writers[:temperature] = Oceananigans.OutputWriters.JLD2Writer(
    integrator,
    (; temperature = integrator.state.temperature);
    filename = "soil_heat_output.jld2",
    schedule = TimeInterval(Δt),
    overwrite_existing = true
)

run!(simulation)
````
and it seems to work great! I had to tweak some of the other code so let me know if this implementation is what you had in mind. Happy to adjust!

(also for future reference `JLD2OutputWriter` is not available in the version of Oceananigans that Terrarium is pinned to but I was able to access it by calling  `Oceananigans.OutputWriters.JLD2Writer` directly)

elenamiralles:
Yes, it was the initial installation! I also was surprised by how long it took. It could have been my internet connection but I've not had that problem with other packages yet.

bgroenks96:
Hey @elenamiralles, sorry for the long silence. The last few weeks were a bit hectic with a lot of traveling, but I'm back now. I think this looks generally good. I updated the branch with the latest changes from main and merged your example script with the existing soil heat column example in [65dd159](https://github.com/NumericalEarth/Terrarium.jl/pull/103/commits/65dd1598900e7ae9b8c9e5d3321c4b7790142a60). I think this makes sense because they are anyway the same basic model setup.

Is there anything else you'd like to add here? One thing that might be nice is testing against the analytical solution to the [Stefan problem](https://github.com/CryoGrid/CryoGrid.jl/blob/master/src/Physics/Heat/analytic/stefan_analytic.jl), but we don't necessarily have to do that in this PR.

bgroenks96 (approved):

elenamiralles:
Hey @bgroenks96!

No worries, I figured that was the case. Glad it looked good! I'd be happy to test it against the Stephan problem too if you think that would be helpful. I'll try to brainstorm some more ideas and let you know if I come up with anything.

PR #104, opened 2026-04-14
Add NF specific intialisation Heun

olivierbonte:
For `ForwardEuler`, initialisation with the `NF` is possible, e.g. `ForwardEuler(Float32)` as mentioned in [the docs](https://numericalearth.github.io/Terrarium.jl/v0/running/time_stepping/#Overview).

For `Heun` however, `Heun(NF)` fails because a method is missing. This PR adds the method needed to allow `Heun(NF)`.

### DIFF
--- a/src/timesteppers/heun.jl
+++ b/src/timesteppers/heun.jl
@@ -11,6 +11,8 @@ Simple forward 2nd order Heun / improved Euler time stepping scheme.
     stage::Stage = nothing
 end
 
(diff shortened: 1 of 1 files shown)


bgroenks96 (approved):

bgroenks96:
Ah right, I noticed this a few weeks ago and then forgot to fix it :man_facepalming:. Thanks!

Issue #106, opened 2026-04-17
Wrong/unconventional use of term of vapor pressure deficit


olivierbonte:
In the current documentation and code, vapor pressure deficit is defined as the difference in vapor pressure between the surface and the atmosphere. See e.g. https://github.com/NumericalEarth/Terrarium.jl/blob/ca59c6018a7d993032b665b55a2cfa25d6cbe787/docs/src/processes/atmosphere/atmosphere.md?plain=1#L45-L60

This does not correspond with the conventional definition (see e.g. [Wikipedia](https://en.wikipedia.org/wiki/Vapour-pressure_deficit)) that the VPD is the difference between actual vapour pressure and saturated at vapour pressure at a certain atmospheric temperature.

What is now called VPD, I would rename to something like "vapor pressure difference".

bgroenks96:
It's not wrong. Vapor pressure varies based on the temperature of the surface. See [Alduchov and Eskridge (1997)](https://www.osti.gov/servlets/purl/548871). The implementation in Terrarium is based on that of CryoGrid.

bgroenks96:
> vapor pressure deficit is defined as the difference in vapor pressure between the surface and the atmosphere.

That's also not what it says... $e_{\text{sat}}$ is still the vapor pressure at saturation, over a surface at the given temperature.

Whether or not that temperature should be the skin temperature is a different question... in CryoGrid it's actually the soil temperature.

olivierbonte:
I am not sure if I am understanding correctly. 
I defenitely do agree with the definition of the saturation vapour pressure. The formulation as in cyrogrid 

<img width="353" height="288" alt="Image" src="https://github.com/user-attachments/assets/8cdd6e58-6b6d-4560-b40d-67cdbc1c1cdd" />

is defenitely a valid option.

To get the saturated vapour pressure at the surface, I also do agree that you should use a surface (or ground temperature) as an input in to the equation B3 in the picture.

What I do not agree with is that the term VPD is correct to quantify the difference between $`e^*(T_s)`$ and $`e(T_a)`$. In Terrarimum, VPD is defined as:
$$VPD = e^*(T_s) - e(T_a)$$

This is however not the definition of VPD as far as I understand. VPD is defined at one temperature (see e.g. [the wiki](https://en.wikipedia.org/wiki/Vapour-pressure_deficit)):
$$VPD(T_a) = e^*(T_a) - e(T_a)$$

[...]

PR #108, opened 2026-04-21
Fixing use of vapor pressure deficit

olivierbonte:
First attempt at closing https://github.com/NumericalEarth/Terrarium.jl/issues/106

2 questions before continuing @bgroenks96 :

1. Is the VPD functionality a physics util or something that is part of the atmosphere? Right now, docs are in atmosphere but the core computation in physics utils, which is a bit confusing.
2. Instead of implementing a lot of the basic functionality in physics_utils, have you considered using [Thermodynamics.jl](https://github.com/CliMA/Thermodynamics.jl)? Just as for the surface aerodynamics (https://github.com/NumericalEarth/Terrarium.jl/issues/97), maybe there is an for wanting the thermodynamics to be consistent across multiple earth system components. This package would give functionality for e.g. calculating VPD (see [here](https://clima.github.io/Thermodynamics.jl/v1.2/API/#Thermodynamics.vapor_pressure_deficit) and saturation vapor pressure (see [here](https://clima.github.io/Thermodynamics.jl/v1.2/API/#Thermodynamics.saturation_vapor_pressure)). It is already a dependency of NumericalEarth.jl (see [here](https://github.com/NumericalEarth/NumericalEarth.jl/blob/6ec0e7714525e642b22799c817a72af7718c40c1/Project.toml#L30)), although I don't know to which extent it is used in that project. If interested, I'd be happy to help wit the PR for this.

### DIFF
--- a/src/processes/atmosphere/prescribed_atmosphere.jl
+++ b/src/processes/atmosphere/prescribed_atmosphere.jl
@@ -180,7 +180,7 @@ Retrieve or compute the specific_humidity at the current time step.
 
 Computes the vapor pressure deficit (VPD) at atmospheric reference level given the current atmospheric fields
 """
-@propagate_inbounds function compute_vapor_pressure_deficit(i, j, grid, fields, atmos::AbstractAtmosphere, c::PhysicalConstants, Ts = nothing)
+@propagate_inbounds function compute_vapor_pressure_deficit(i, j, grid, fields, atmos::AbstractAtmosphere, c::PhysicalConstants)
     T_air = air_temperature(i, j, grid, fields, atmos)
     q_air = specific_humidity(i, j, grid, fields, atmos)
     p = air_pressure(i, j, grid, fields, atmos)
--- a/src/processes/atmosphere/prescribed_atmosphere.jl
+++ b/src/processes/atmosphere/prescribed_atmosphere.jl
@@ -187,32 +187,6 @@ Computes the vapor pressure deficit (VPD) at atmospheric reference level given t
     vpd = vapor_pressure_deficit(c, p, q_air, T_air)
     return vpd
 end
-"""
-    $TYPEDSIGNATURES
-
-Computes the specific humidity difference between the surface at temperature `Ts` and the current atmospheric fields.
-"""
-@propagate_inbounds function compute_specific_humidity_difference(i, j, grid, fields, atmos::AbstractAtmosphere, c::PhysicalConstants, Ts)
-    let Δe = compute_vapor_pressure_difference(i, j, grid, fields, atmos, c, Ts),
-            p = air_pressure(i, j, grid, fields, atmos)
-        Δq = vapor_pressure_to_specific_humidity(Δe, p, c.ε)
-        return Δq
-    end
-end
-
-"""
-    $TYPEDSIGNATURES
-
-Computes the vapor pressure difference between a surface at temperature `Ts` and the current atmospheric fields.
-"""
-@propagate_inbounds function compute_vapor_pressure_difference(i, j, grid, fields, atmos::AbstractAtmosphere, c::PhysicalConstants, Ts)
-    Tair = air_temperature(i, j, grid, fields, atmos)
-    q_air = specific_humidity(i, j, grid, fields, atmos)
-    pres = air_pressure(i, j, grid, fields, atmos)
-    e_air = specific_humidity_to_vapor_pressure(q_air, pres, c.ε)
-    e_sat_s = saturation_vapor_pressure(Ts)
-    return e_sat_s
-end
 
 """
     $TYPEDEF
--- a/src/processes/atmosphere/prescribed_atmosphere.jl
+++ b/src/processes/atmosphere/prescribed_atmosphere.jl
@@ -164,7 +164,7 @@ Computes the vapor pressure deficit (VPD) at atmospheric reference level given t
     T_air = air_temperature(i, j, grid, fields, atmos)
     q_air = specific_humidity(i, j, grid, fields, atmos)
     p = air_pressure(i, j, grid, fields, atmos)
-    vpd = compute_vapor_pressure_deficit(c, p, q_air, T_air)
+    vpd = vapor_pressure_deficit(c, p, q_air, T_air)
     return vpd
 end
 """
--- a/src/processes/atmosphere/prescribed_atmosphere.jl
+++ b/src/processes/atmosphere/prescribed_atmosphere.jl
@@ -103,6 +103,26 @@ variables(atmos::PrescribedAtmosphere{NF}) where {NF} = (
 @inline compute_tendencies!(state, grid, atmos::PrescribedAtmosphere) = nothing
 
 """
+    $SIGNATURES
+
+Computes the vapor pressure deficit for an air parcel at temperature `T` with surface pressure `pres`
+and specific humidity of air `q_air`. Assumes that air parcel is over water when `T > 0°C` and over 
+ice when `T < 0°C`.
+"""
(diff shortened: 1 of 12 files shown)


bgroenks96:
If we follow the discussed conventions, I would make VPD part of the atmosphere but move the vapor pressure difference to the surface energy balance code.

The `saturation_vapor_pressure` and related functions should remain utilities, I guess, but you might need to add a separate dispatch that doesn't use the surface temperature. I don't think the existing one is appropriate for VPD at an arbitrary reference height.

olivierbonte:
Okay, agree that this sort of placement of functions makes sense. I'll update the code accordingly and let you know once it is ready for review.

bgroenks96:
Regarding Thermodynamics.jl, if you don't see any technical or structural barriers to using it, I am ok with it.

bgroenks96:
The main thing to check would be whether it's Enzyme compatible and how/where they store parameters (or rather constants).

bgroenks96:
Ah I think they implicitly require the use of `ClimaParams` (which we definitely do not want) or we would need to define our own parameter sets. That honestly seems like more trouble than it's worth?

Edit: Eh, I don't know, maybe it's not too bad. I'll leave it up to you to decide.

olivierbonte:
@bgroenks96 I now have a version (without relying on Thermodynamics.jl) that is ready for review. 

I think there are still advantages to using Thermodynamics.jl, but I'd rather implement this in a separate PR. In NumericalEarth.jl, they subtype `AbstractThermodynamicsParameters` (from Thermodynamics.jl), which is I think is an interesting approach: https://github.com/NumericalEarth/NumericalEarth.jl/blob/a5ab40014911f10b627047c0b80d9af0f8334e3c/src/Atmospheres/thermodynamic_parameters.jl

Is there any way we can reuse what they have already implemented? Just as for the surface fluxes, I think NumericalEarth.jl has a lot of interesting stuff already implemented, but not sure how this can be used within Terrarium.jl, as I suppose you don't want it as a direct dependency?

src/processes/atmosphere/prescribed_atmosphere.jl
```diff
@@ -158,27 +178,14 @@ Retrieve or compute the specific_humidity at the current time step.
...
-
-"""
-    $TYPEDSIGNATURES
-
-Computes the vapor pressure deficit over a surface at temperature `Ts` from the current atmospheric fields.
+Computes the vapor pressure deficit (VPD) at atmospheric reference level given the current atmospheric fields
 """
-@propagate_inbounds function compute_vpd(i, j, grid, fields, atmos::AbstractAtmosphere, c::PhysicalConstants, Ts = nothing)
-    Tair = air_temperature(i, j, grid, fields, atmos)
+@propagate_inbounds function compute_vapor_pressure_deficit(i, j, grid, fields, atmos::AbstractAtmosphere, c::PhysicalConstants, Ts = nothing)
```
bgroenks96: You can remove the `Ts` argument.

src/processes/surface_energy/turbulent_fluxes.jl
```diff
@@ -49,6 +49,32 @@ function compute_latent_heat_flux(::DiagnosedTurbulentFluxes, Q_h, ρₐ, Lsl)
     return Hₗ
 end
 
+"""
+    $TYPEDSIGNATURES
+    
+Computes the difference in vapor pressure between the (saturated) surface at temperature `Ts`
+and the atmosphere, defined by its specific humidity `q_air`.
+"""
+function vapor_pressure_difference(c::PhysicalConstants, p, q_air, Ts)
```
bgroenks96: Can we maybe call these `surface_*_difference` or `*_difference_at_surface` to be extra clear?

src/processes/physics_utils.jl
```diff
@@ -54,12 +65,11 @@ coefficients a₁, a₂, and a₃.
...
 
 """
-    saturation_vapor_pressure(T, Ts=T)
-
-Saturation vapor pressure at the given temperature `T`, accounting for both frozen (`T < 0°C`)
-and unfrozen conditions.
+    saturation_vapor_pressure(T)
 
-Coefficients taken from [alduchovImprovedMagnusForm1996](@cite).
+Saturation vapor pressure of an air parcel at the given temperature `T`. By default, the saturation vapor
```
bgroenks96, line 70: Are we sure that this is valid for VPD at arbitrary reference height? I thought the experiments in this paper specifically looked at surface vapor pressure.

bgroenks96 (changes requested):

bgroenks96:
> Is there any way we can reuse what they have already implemented? Just as for the surface fluxes, I think NumericalEarth.jl has a lot of interesting stuff already implemented, but not sure how this can be used within Terrarium.jl, as I suppose you don't want it as a direct dependency?

Yeah no it can't be a direct dependency because the dependency is supposed to go the other direction; i.e. `NumericalEarth` will have an extension for `Terrarium`. So we would need to refactor their `Thermodynamics` parameters into a separate package.

src/processes/surface_energy/turbulent_fluxes.jl
olivierbonte: Yes, I think that's a good idea. I'll change to `surface_*_difference`.

olivierbonte: Upon implementing this, I am not 100% sure if specifying surface here always make sense. For example in https://github.com/NumericalEarth/Terrarium.jl/pull/108/changes#diff-6148e7d6a2b3a70f3a221bad0dcf96b1a2fedda194b3416bd6cdd6ccdec7aeb7R146 this surface prefix would be suboptimal, as you are using ground temperature. You could use the same function for a canopy as well, and then again the surface prefix might be confusing?

src/processes/physics_utils.jl
olivierbonte: I briefly checked https://doi.org/10.1175/1520-0450(1996)035%3C0601:IMFAOS%3E2.0.CO;2 again, and as far as I understand, this relationship should be valid at any height. They specifically mention the use case of an upper air dataset. 

These relationships are derived with respect to a plane surface (ice or water, see e.g. https://clima.github.io/Thermodynamics.jl/dev/Formulation/#9.-Saturation-Vapor-Pressure), but this doesn't limit their use to just above the surface. In the referenced docs above, they mention the use of these type of relationships for calculating saturated vapor pressure in clouds.

src/processes/surface_energy/turbulent_fluxes.jl
bgroenks96: Ah, ok, good point. Though this is at least partially a problem with "surface" being ambiguous. But I would at least rename `Ts` to `T` and specify that it can be any temperature.

src/processes/physics_utils.jl
bgroenks96: Ok, thanks for checking!

src/processes/atmosphere/prescribed_atmosphere.jl
olivierbonte: Fixed in https://github.com/NumericalEarth/Terrarium.jl/pull/108/commits/11a35304341e19c8b2252e1d11884e68719bf3fa

src/processes/surface_energy/turbulent_fluxes.jl
olivierbonte: Good point, addressed in https://github.com/NumericalEarth/Terrarium.jl/pull/108/commits/fc43a739c42eb3cc11017b69fa86c91f9fa48a8c

bgroenks96 (approved):

bgroenks96:
Thanks @olivierbonte!

PR #110, opened 2026-04-22
Coupling `LandModel` with full surface energy balance to SpeedyWeather

bgroenks96:
This PR updates the old SpeedyWeather script for the `PrimitiveDryModel` air-soil temperature coupling and adds a new script with prototype coupling to a `PrimitiveWetModel` via the full SEB.

In the process of implementing this, I am also fixing bugs in the surface energy/water flux calculations.

- The ground heat flux calculation was previously using an incorrect sign convention. It should be `G = R_net + H_s + H_l` corresponding to balance equation `R_net + H_s + H_l - G = 0` with all fluxes positive upwards (towards the atmosphere).
- The soil moisture evaporation resistance factor code was broken (we should add more tests...)
- `SoilMoistureResistanceFactor` is now the default choice for `BareGroundEvaporation`.
- Default values for incoming shortwave and longwave radiation now match the standard values for the global energy balance


examples/simulations/speedy_wet_land.jl
```diff
@@ -0,0 +1,175 @@
...
+    set!(state.inputs.windspeed, wind)
+    set!(state.inputs.surface_shortwave_down, SwIn)
+    set!(state.inputs.surface_longwave_down, LwIn)
+    # run land forward over speedy timestep interval;
+    # we use a smaller actual timestep to ensure stability
+    Terrarium.run!(land.integrator, period = progn.clock.Δt, Δt = 60.0)
+    # Update speedy variables
+    progn.land.soil_temperature .= state.skin_temperature .+ NF(273.15)
+    progn.land.sensible_heat_flux .= state.sensible_heat_flux
+    progn.land.surface_humidity_flux .= state.latent_heat_flux ./ consts.Lsl
```
bgroenks96: @maximilian-gelbrecht and @milankl is this correct? or do I need to multiply by the density of air to cancel out the specific humidity units?

examples/Project.toml
maximilian-gelbrecht, line 1: Maybe good to add a compat that restricts to 0.18.1 so that it doesn't directly break once we release the new version

examples/simulations/speedy_dry_land.jl
```diff
@@ -86,30 +98,26 @@ integrator = initialize(model, ForwardEuler(eltype(grid)), Tair_input, boundary_
...
 srad_fig = heatmap(exp.(sim_coupled.diagnostic_variables.physics.surface_shortwave_down), title = "Surface shortwave down", size = (800, 400))
-# Tskin_fig = heatmap(RingGrids.Field(interior(integrator.state.skin_temperature)[:,1,end], grid), title="", size=(800,400))
-# save("plots/speedy_primitive_dry_coupled_tair_R48.png", Tair_fig, px_per_unit=1)
-# save("plots/speedy_primitive_dry_coupled_tsoil_R48.png", Tsoil_fig, px_per_unit=1)
 
 # animate surface air and soil temperatures
-Speedy.animate(sim, variable = "temp", coastlines = false, level = spectral_grid.nlayers, output_file = "plots/speedy_terrarium_dry_air_temperature.mp4")
-Speedy.animate(sim, variable = "st", coastlines = false, level = 1, output_file = "plots/speedy_terrarium_dry_soil_temperature.mp4")
+Speedy.animate(sim_coupled, variable = "temp", coastlines = false, level = spectral_grid.nlayers, output_file = "plots/speedy_terrarium_dry_air_temperature.mp4")
+# Speedy.animate(sim_coupled, variable = "st", coastlines = false, level = 1, output_file = "plots/speedy_terrarium_dry_soil_temperature.mp4") # TODO: this is broken now for some reason?
```
maximilian-gelbrecht, line 120: What error do you get?

examples/simulations/speedy_wet_land.jl
maximilian-gelbrecht: I am fairly certain this is correct. At least the units work out 😅

examples/simulations/speedy_wet_land.jl
```diff
@@ -0,0 +1,175 @@
...
+    humid = @view diagn.grid.humid_grid[:, end]
+    pres = diagn.grid.pres_grid
+    wind = diagn.physics.surface_wind_speed
+    rain = diagn.physics.rain_rate
+    snow = diagn.physics.snow_rate
+    SwIn = diagn.physics.surface_shortwave_down
+    LwIn = diagn.physics.surface_longwave_down
+    # update Terrarium input fields
+    state = land.integrator.state
+    set!(state.inputs.air_temperature, Tair) # first set directly to avoid allocating new fields
```
maximilian-gelbrecht, line 90: why is this needed?

src/models/soil/soil_model_init.jl
```diff
@@ -140,13 +140,13 @@ Properties:
...
-    vadose_zone_saturation::NF = 0.5
+    vadose_zone_saturation::NF = 0.75
     water_table_depth::NF = 5.0
 end
 
 SaturationWaterTable(::Type{NF}; kwargs...) where {NF} = SaturationWaterTable{NF}(; kwargs...)
 
 function initialize!(state, ::AbstractModel, init::SaturationWaterTable{NF}) where {NF}
-    set!(state.saturation_water_ice, (x, z) -> z <= init.water_table_depth ? one(NF) : init.vadose_zone_saturation)
+    set!(state.saturation_water_ice, (x, z) -> z <= -init.water_table_depth ? one(NF) : init.vadose_zone_saturation)
```
maximilian-gelbrecht, line 150: What's the orientation of the water table now? And what units is this?

bgroenks96: It doesn't have an orientation, it's just an elevation with coordinates on the Z-axis. The code here was just wrong. Probably the field name should be `water_table_elevation` though...

examples/simulations/speedy_wet_land.jl
bgroenks96: as the comment said, the point is to avoid allocations.

examples/simulations/speedy_wet_land.jl
bgroenks96: `Lsl` is wrong though... should be `Llg`.

examples/simulations/speedy_wet_land.jl
maximilian-gelbrecht: I am just wondering why the line below doesn't cause allocations with `state.inputs.air_temperature` instead of `T_air`. Does `set!` compute this differently in this case?

examples/simulations/speedy_wet_land.jl
milankl: What is `Lsl`, `Llg`? We also started off using "accepted" nomenclature like `L` for latent heat but the problem is always that there's sooo many sciences being put together when you build a model like this and then an acronym is either inaccurate, cryptic or both. Why not write the full thing and if there's an equation where you'd like to write it brief just rename locally?

But yes, if this is the latent heat flux in W/m^2 then you need to divide by latent heat of vaporization to get to kg/m^2/s for the humidity flux. To force our land model we do the opposite: humidity flux * latent heat 

https://github.com/SpeedyWeather/SpeedyWeather.jl/blob/64f560affd1ecee9ee172883606fd9e73d63f8dd/SpeedyWeather/src/parameterizations/land/soil_temperature.jl#L278-L282

milankl: Do you already deal with sublimation? Because then the latent heats are different ... in which case you would need to keep track of two latent heat fluxes, or internally convert directly to the humidity flux?

examples/simulations/speedy_dry_land.jl
milankl: There's still https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/958 as an open pull request ...

examples/simulations/speedy_wet_land.jl
bgroenks96: because `state.inputs.air_temperature` is an Oceananigans `Field` that supports `AbstractOperation`s, so `exp` is lazily evaluated by `set!`.

examples/simulations/speedy_wet_land.jl
bgroenks96: `Lsl` is latent heat solid -> liquid (fusion) and `Llg` is liquid -> gas (vaporization). Sublimation is `Lsg`, solid -> gas.

We don't deal with sublimation yet, but yes, I agree that the best solution here would be to just make `surface_humidity_flux` a state variable so the model can deal with this internally rather in the coupling code.

bgroenks96: I also agree that the fields in `PhysicalConstants` should be explicitly spelled out, will change that in a future PR.

examples/Project.toml
bgroenks96: Done in [6c51fc6](https://github.com/NumericalEarth/Terrarium.jl/pull/110/commits/6c51fc6ebb4503db3c6b87dac2ffbb2aed50b329)

examples/simulations/speedy_dry_land.jl
bgroenks96: These lines are now removed.

bgroenks96:
@milankl Since @maximilian-gelbrecht is on vacation, it would be great if you could approve this PR.

bgroenks96:
@maximilian-gelbrecht and @milankl I have verified that the coupled model runs for up to 1 month of simulated time without errors, though I did not check the output. This is without the Richard's equation soil hydrology since this currently has problems unless you reduce the step size to an unbearably small number (like <10 seconds). This is, to some extent, expected; Richards' equation really needs implicit timestepping to be efficient.

examples/simulations/speedy_dry_land.jl
```diff
@@ -14,38 +14,52 @@ import SpeedyWeather as Speedy
...
+
 function Speedy.initialize!(
-        progn_land::Speedy.PrognosticVariablesLand,  # for dispatch
-        progn::Speedy.PrognosticVariables{NF},
-        diagn::Speedy.DiagnosticVariables{NF},
-        land::TerrariumDryLand,
+        ::Any,
+        progn::Speedy.PrognosticVariables,
+        diagn::Speedy.DiagnosticVariables,
+        land::TerrariumDryLand{NF},
```
milankl, line 50: heads up, in v0.19 this signature changed to

```julia
::Variables, ::SpeedyWeather.AbstractLandModel, ::SpeedyWeather.AbstractModel
```

examples/simulations/speedy_dry_land.jl
```diff
@@ -14,38 +14,52 @@ import SpeedyWeather as Speedy
...
     return nothing
 end
 
 function Speedy.timestep!(
-        progn::Speedy.PrognosticVariables{NF},
-        diagn::Speedy.DiagnosticVariables{NF},
-        land::TerrariumDryLand,
+        progn::Speedy.PrognosticVariables,
+        diagn::Speedy.DiagnosticVariables,
+        land::TerrariumDryLand{NF},
```
milankl, line 62: Same here

src/processes/atmosphere/prescribed_atmosphere.jl
```diff
@@ -1,23 +1,24 @@
...
 )
 
-default_tracer_conc(::Val{:CO2}) = 380 # in ppm
-
 """
 Creates a `TracerGas` for ambient CO2 with concentration prescribed by an input variable with
 the given name.
 """
-AmbientCO2(name = :CO2) = TracerGas(name)
+AmbientCO2(::Type{NF}, name = :CO2) where {NF} = TracerGas(name, NF(380))
```
milankl, line 21: Change to our greenhouse gas interface once that's merged? https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/993

src/processes/surface_energy/turbulent_fluxes.jl
```diff
@@ -89,12 +89,12 @@ Compute the sensible heat flux at `i, j` based on the current skin temperature a
...
         atmos::AbstractAtmosphere
     )
-    let ρₐ = constants.ρₐ, # density of air
-            cₐ = constants.cₐ, # specific heat capacity of air
-            rₐ = aerodynamic_resistance(i, j, grid, fields, atmos), # aerodynamic resistance
-            Tₛ = skin_temperature(i, j, grid, fields, skinT), # skin temperature
-            Tₐ = air_temperature(i, j, grid, fields, atmos), # air temperature
-            Q_T = (Tₛ - Tₐ) / rₐ  # bulk aerodynamic temperature-gradient
+    let ρₐ = constants.ρₐ # density of air
+        cₐ = constants.cₐ # specific heat capacity of air
```
milankl, line 143: Can you not name the `constants` as written in the comments? 
```julia
ρₐ = constants.air_density
```
or similar? That means you're just locally renaming to a brief math symbol and one doesn't need the additional comment to understand what's going on

examples/simulations/speedy_wet_land.jl
```diff
@@ -0,0 +1,177 @@
...
+        spectral_grid = Speedy.SpectralGrid(integrator.model.grid.rings; NF, spectral_grid_kwargs...)
+        land_grid = integrator.grid
+        Δz = on_architecture(CPU(), land_grid.z.Δᵃᵃᶜ)
+        geometry = Speedy.LandGeometry(1, Δz[end])
+        return new{eltype(integrator), typeof(geometry), typeof(integrator)}(spectral_grid, geometry, integrator)
+    end
+end
+
+Speedy.variables(land::TerrariumWetLand) = (
+    Speedy.PrognosticVariable(name = :soil_temperature, dims = Speedy.Grid3D(), namespace = :land),
```
milankl, line 43: Ouh why defining a `<: AbstractWetLand` and then not having `soil_moisture` as variable or component?

milankl (commented):
Just some comments from my side skimming through!!!

src/processes/surface_energy/turbulent_fluxes.jl
bgroenks96: as discussed in the other thread, we will do this in another PR. See #111.

src/processes/atmosphere/prescribed_atmosphere.jl
bgroenks96: Terrarium does not depend directly on `SpeedyWeather`, so I don't think that would be possible. We could align the names maybe.

examples/simulations/speedy_wet_land.jl
bgroenks96: Is that documented somewhere?

bgroenks96: Fixed by [05cb25c](https://github.com/NumericalEarth/Terrarium.jl/pull/110/commits/05cb25c6c4207a06f23eb3bdb75a526daabf374e)

examples/simulations/speedy_dry_land.jl
bgroenks96: I added a compat bound to temporarily put off dealing with this.

maximilian-gelbrecht:
Do you want to merge this PR first, and then in a follow up adjust to Speedy 0.19, or still do that here? I am also happy to help with that adjustment.

bgroenks96:
#114 already updates to Speedy v0.19

maximilian-gelbrecht:
Yesterday(?) we briefly discussed that it's actually not that nice that with the current setup the Terrarium variables are hidden in the model struct. 

I think there could be a fix. Maybe it's not the long-term solution but we could basically frankenstein both variable containers together. 

In the coupling script / Terrarium define 

```Julia
struct TerrariumVars <: SpeedyWeather.AbstractVariableDim end 
Base.zero(::TerrariumVars, model::SpeedyWeather.AbstractModel) = Terrarium.initialize(model.land.model, clock=model.land.clock, boundary_conditions=model.land.boundary_conditions, input_variables=model.land.input_variables, fields = model.land.fields)

SpeedyWeather.variables(land::TerrariumWetLand) = (
    SpeedyWeather.PrognosticVariable(name = :terrarium, dims = TerrariumVars(), namespace = :land)
)
```
This would just allocate the Terrarium `StateVars` within the Speedy `Variables`. That should work in principle as far as I can see it. 

What do you think @bgroenks96? 

That should work right? One just needs a slightly modified `initialize` for the `ModelIntegrator` on Terrarium side. Or am I seeing this wrong?

bgroenks96:
Hey @maximilian-gelbrecht, I think this could work but a couple of things:

```julia
Base.zero(::TerrariumVars, model::SpeedyWeather.AbstractModel) = Terrarium.initialize(model.land.model, clock=model.land.clock, boundary_conditions=model.land.boundary_conditions, input_variables=model.land.input_variables, fields = model.land.fields)
```
As discussed in person, I think this would be an abuse of the `Base.zero` API. Terrarium's `initialize` is not generally guaranteed to return zero-valued `Field`s, though I guess you could rectify this by adding `|> zero` since I think `StateVariables` defines `zero`.

I don't understand what this part is doing:
```julia
SpeedyWeather.variables(land::TerrariumWetLand) = (
    SpeedyWeather.PrognosticVariable(name = :terrarium, dims = TerrariumVars(), namespace = :land)
)
```
is this meant to be an example? Or is it literally creating a prognostic variable that has type `TerrariumVars`? That seems rather unintuitive given that here its assigned to `dims`. But maybe I just don't understand how Speedy's variable system works.

In any case, I think this discussion should be continued in a separate issue/PR, no? It's not really necessary for this one.

maximilian-gelbrecht:
It's not meant as an example, this is how SpeedyWeather allocates variables. It is creating a prognostic variable that has the same type as the return of `zero(::TerrariumVars, model)`

Aside from the clock all other variable type just declare arrays so `zero` did feel okay initially. I am open to renaming things, I guess `allocate` would be more general, but do you have any concerns aside from naming choices in SpeedyWeather?

bgroenks96:
> Aside from the clock all other variable type just declare arrays so zero did feel okay initially. 

That might be true in SpeedyWeather but it's not in Terrarium. `initialize` invokes the model initializer which may set state variables to any arbitrary values as defined by the initializers.

> do you have any concerns aside from naming choices in SpeedyWeather?

I don't think it's a long-term solution because it's very strange to treat the entire Terrarium `StateVariables` as a single prognostic variable in SpeedyWeather; but it seems to be a reasonable temporary Frankenstein solution.

maximilian-gelbrecht (approved):

Issue #111, opened 2026-04-27
Use spelled out names in `PhysicalConstants`


bgroenks96:
From discussion in #110 and related to #109.

olivierbonte:
Resolved in #112

PR #112, opened 2026-04-30
Adding Clima Thermodynamics

olivierbonte:
As previously mentioned in https://github.com/NumericalEarth/Terrarium.jl/pull/108, I thought Thermodynamics.jl could be a nice way to avoid reimplementation of a lot of the thermodynamics. This PR is an attempt at replacing some utilities by functions from Thermodynamics.jl.  

To avoid the use of [ClimaParams.jl](https://github.com/NumericalEarth/Terrarium.jl/pull/108), `PhysicalConstants` is defined as a subtype of [`AbstractThermodynamicsParameters`](https://github.com/CliMA/Thermodynamics.jl/blob/e19d95dcafd07f1995df586f76db4e6c11579392/src/Parameters.jl#L10). This approach is inspired by the use of Thermodynamics.jl in NumericalEarth.jl for the atmospheric thermodynamics (see [here](https://github.com/NumericalEarth/NumericalEarth.jl/blob/0d399ae125f9e1786e34667c13454bd932b18465/src/Atmospheres/thermodynamic_parameters.jl)). 

This PR also introduces specific heat capacity and air density as functions, not constants, consistent with the use in Thermodynamics.jl


olivierbonte:
I now also see that there is an issue on the use of abbreviations in physical constants https://github.com/NumericalEarth/Terrarium.jl/issues/111. Is this something to fix in this PR or not @bgroenks96 ?

bgroenks96:
Hey @olivierbonte, I think this PR would be a good opportunity to resolve #111!

I had also considered splitting up PhysicalConstants into separate types for thermodynamic vs material and hydraulic constants. What do you think about this?

olivierbonte:
Alright! I'll try to address this soon 

I agree that splitting up the constants in categories makes sense, otherwise this one struct will get very large I think. Given the current set of properties, thermodynamic vs material makes sense to me. Any preferences for the naming of these structs?

src/processes/physical_constants.jl
```diff
@@ -6,58 +22,98 @@ A collection of general physical constants that do not (usually) need to be vari
...
 
+@inline R_d(c::PhysicalConstants) = c.R_d
+@inline R_v(c::PhysicalConstants) = c.R_v
+@inline cp_d(c::PhysicalConstants) = c.cp_d
+@inline cp_i(c::PhysicalConstants) = c.cp_i
+@inline cp_l(c::PhysicalConstants) = c.cp_l
+@inline cp_v(c::PhysicalConstants) = c.cp_v
+@inline LH_v0(c::PhysicalConstants) = c.Llg
+@inline LH_s0(c::PhysicalConstants) = c.Lsg
+@inline T_0(c::PhysicalConstants) = c.T_ref
```
bgroenks96: Why is this `T_0` and not `T_ref`? I guess you can also address this when updating the names to be fully spelled out.

src/processes/physics_utils.jl
```diff
@@ -30,54 +38,56 @@ Compute partial pressure of CO2 from surface pressure and CO2 concentration in P
...
-August-Roche-Magnus equation for saturation vapor pressure at temperature `T` with empirical
-coefficients a₁, a₂, and a₃.
+Saturation vapor pressure of an air parcel at the given temperature `T` in °C. By default, the saturation vapor
+pressure is computed over ice for `T <= 0°C` and over water for `T > 0°C`. Wrapper around
+[`saturation_vapor_pressure`](@extref Thermodynamics.saturation_vapor_pressure).
 """
-@inline saturation_vapor_pressure(T, a₁, a₂, a₃) = a₁ * exp(a₂ * T / (T + a₃))
+@inline function saturation_vapor_pressure(c::PhysicalConstants, T::NF) where {NF}
+    T_K = celsius_to_kelvin(c, T)
+    return if T <= zero(T)
```
bgroenks96: I know this also isn't what the old code did, but maybe we should use `ifelse` here just to be safe (for Enzyme/GPU)?

bgroenks96 (changes requested):
Looks really great! Thanks a lot!

bgroenks96:
@olivierbonte Any updates on this? I guess once we merge this we could release a v0.1.1.

olivierbonte:
Sorry for the delay on this @bgroenks96, had some deadlines for teaching responsibilities last week, but this week I'll pick this up again!

src/processes/physical_constants.jl
olivierbonte: The reason is that you need to explicitly define a new method `T_0` based on the exported method `T_0` from thermodynamics.jl for the code to work. So in `Thermodynamics.jl`, with `p` the param set they call `T_0(p)` to access `T_0` (`T_ref` in our case).

Should this be made more clear in the file that this is the goal of all these inline functions?

bgroenks96: Ah, ok, I didn't notice that. In that case, I think you should explicitly prefix these method definitions to be clear where they came from.

src/processes/physical_constants.jl
```diff
@@ -1,3 +1,19 @@
+import Thermodynamics.Parameters:
```
bgroenks96: Right, so we generally try to avoid doing this. I would rather do `import Thermodynamics` in the top-level module and then explicitly prefix everything here, as mentioned in the other comment.

bgroenks96: Sorry for missing this the first time.

src/processes/physical_constants.jl
olivierbonte: By this you mean using `Thermodynamics.T_0` instead of just `T_0` (for example)?

bgroenks96 (changes requested):
Oops, sorry forgot to submit these comments yesterday. I always find this "review" feature of GitHub a bit unintuitive.

src/processes/physical_constants.jl
olivierbonte: okay perfect, by top-level module you mean doing the import in the Terrarium.jl file?

bgroenks96: Yes!

olivierbonte:
@bgroenks96 I think all comments are addressed now, let me know what you think!

docs/make.jl
```diff
@@ -146,8 +146,12 @@ links = InterLinks(
     "KernelAbstractions" => "https://juliagpu.github.io/KernelAbstractions.jl/stable/",
     "SpeedyWeather" => "https://speedyweather.github.io/SpeedyWeatherDocumentation/stable/",
     "FreezeCurves" => "https://cryogrid.github.io/FreezeCurves.jl/stable/",
+    "Thermodynamics" => "https://clima.github.io/Thermodynamics.jl/stable/",
 )
 
+# Always have `using Terrarium` available in doctests
+DocMeta.setdocmeta!(Terrarium, :DocTestSetup, :(using Terrarium); recursive = true)
```
bgroenks96, line 153: Nice, didn't know about this!

docs/src/processes/utils/physical_constants.md
```diff
@@ -14,38 +14,113 @@ using InteractiveUtils
...
-| `σ` | $\sigma$ | 5.6704×10⁻⁸ | W/(m²·K⁴) | Stefan-Boltzmann constant |
-| `κ` | $\kappa$ | 0.4 | — | von Kármán constant |
-| `ε` | $\varepsilon$ | 0.622 | — | Ratio of molecular weights $M_v / M_d$ |
-| `Rₐ` | $R_a$ | 287.058 | J/(kg·K) | Specific gas constant of dry air |
-| `C_mass` | — | 12.0 | gC/mol | Molar mass of carbon |
+| `specific_heat_capacity_dry_air` | $c_{p,d}$ | 1004.5 | J/(kg·K) | Isobaric specific heat capacity of dry air at 0°C |
+| `specific_heat_capacity_ice` | $c_{p,i}$ | 2070.0 | J/(kg·K) | Isobaric specific heat capacity of ice at 0°C |
+| `specific_heat_capacity_liquid_water` | $c_{p,l}$ | 4186.0 | J/(kg·K) | Isobaric specific heat capacity of liquid water at 0°C |
+| `specific_heat_capacity_water_vapor` | $c_{p,v}$ | 1846.0 | J/(kg·K) | Isobaric specific heat capacity of water vapor at 0°C |
+| `latent_heat_fusion_at_reference` | $L_{sl}$ | 3.34×10⁵ | J/kg | Specific latent heat of fusion at 0°C |
```
bgroenks96: Hmm, I get the inclination towards being precise by including `_at_reference` but I wonder whether or not this is really necessary? Can you imagine a situation where we would add more latent heat constants that would *not* be at a (or rather the) reference temperature?

docs/make.jl
maximilian-gelbrecht: What exactly does this do? I also haven't seen it.

src/processes/physical_constants.jl
```diff
@@ -1,75 +1,191 @@
...
+
+julia> c.thermodynamic_constants.temperature_reference
+273.15
+```
+
+Properties:
+$FIELDS
+"""
+struct PhysicalConstants{NF}
+    thermodynamic_constants::ThermodynamicConstants{NF}
```
bgroenks96: To be slightly less verbose, how about `thermodynamics`, `material`, and `unviersal`? That way it's `constants.thermodynamics.specific_heat_capacity_dry_air`. That would read a bit nicer, no?

src/processes/physical_constants.jl
```diff
@@ -1,75 +1,191 @@
...
+@inline Thermodynamics.Parameters.LH_v0(c::ThermodynamicConstants) = c.latent_heat_vaporization_at_reference
+@inline Thermodynamics.Parameters.LH_s0(c::ThermodynamicConstants) = c.latent_heat_sublimation_at_reference
+@inline Thermodynamics.Parameters.T_0(c::ThermodynamicConstants) = c.temperature_reference
+@inline Thermodynamics.Parameters.T_freeze(c::ThermodynamicConstants) = c.temperature_water_freeze
+@inline Thermodynamics.Parameters.T_triple(c::ThermodynamicConstants) = c.temperature_water_triple_point
+@inline Thermodynamics.Parameters.press_triple(c::ThermodynamicConstants) = c.pressure_water_triple_point
+
+# Derived parameters
+@inline Thermodynamics.Parameters.Rv_over_Rd(c::ThermodynamicConstants) = Thermodynamics.Parameters.R_v(c) / Thermodynamics.Parameters.R_d(c)
+@inline ε(c::ThermodynamicConstants) = R_d(c) / R_v(c)
```
bgroenks96: Can you make this explicitly spelled out? I guess it would be `ratio_gas_constant_dry_air_to_water_vapor`? A bit long but I guess OK?

bgroenks96: particularly important given that epsilon (not \varpesilon but still) is used for emissivity.

src/processes/surface_energy/turbulent_fluxes.jl
```diff
@@ -139,12 +144,15 @@ Compute the sensible heat flux at `i, j` based on the current skin temperature a
         constants::PhysicalConstants,
         atmos::AbstractAtmosphere
     )
-    let ρₐ = constants.ρₐ # density of air
-        cₐ = constants.cₐ # specific heat capacity of air
-        rₐ = aerodynamic_resistance(i, j, grid, fields, atmos) # aerodynamic resistance
-        Tₛ = skin_temperature(i, j, grid, fields, skinT) # skin temperature
-        Tₐ = air_temperature(i, j, grid, fields, atmos) # air temperature
-        Q_T = (Tₛ - Tₐ) / rₐ  # bulk aerodynamic temperature-gradient
+    let rₐ = aerodynamic_resistance(i, j, grid, fields, atmos), # aerodynamic resistance
```
bgroenks96: Can you revert the formatting change here? With Runic, I find it better not to use commas in the `let` block because otherwise it does this weird indenting here. We could also disable the formatter on all let blocks, but I think omitting commas is a cleaner solution. They aren't really necessary anyway.

bgroenks96: Actually you could also just remove the let block here since it's the only code in this function. I have a habit of using `let` blocks a lot, but it's admittedly kind of pointless in cases like this.

src/processes/surface_energy/turbulent_fluxes.jl
```diff
@@ -165,12 +173,16 @@ to the latent heat flux.
...
         atmos::AbstractAtmosphere
     )
-    let L = constants.Llg # specific latent heat of vaporization of water
-        ρₐ = constants.ρₐ # density of air
-        Tₛ = skin_temperature(i, j, grid, fields, skinT)
-        rₐ = aerodynamic_resistance(i, j, grid, fields, atmos) # aerodynamic resistance
-        Δq = compute_specific_humidity_difference(i, j, grid, fields, atmos, constants, Tₛ)
-        Q_h = Δq / rₐ  # humidity flux
+    let L = constants.thermodynamic_constants.latent_heat_vaporization_at_reference,
+            Tₛ = skin_temperature(i, j, grid, fields, skinT),
```
bgroenks96: Same thing here

src/processes/surface_energy/turbulent_fluxes.jl
```diff
@@ -188,11 +200,16 @@ defined by `evtr` which is assumed to be already computed.
...
         tur::DiagnosedTurbulentFluxes,
         evtr::AbstractEvapotranspiration,
-        constants::PhysicalConstants
+        constants::PhysicalConstants,
+        atmos::AbstractAtmosphere
     )
-    let L = constants.Llg # specific latent heat of vaporization of water
-        ρₐ = constants.ρₐ # density of air
-        Q_h = surface_humidity_flux(i, j, grid, fields, evtr)   # humidity flux
+    let L = constants.thermodynamic_constants.latent_heat_vaporization_at_reference,
```
bgroenks96: and here

src/processes/vegetation/photosynthesis.jl
```diff
@@ -199,11 +199,11 @@ Follows [willeitPALADYNV10Comprehensive2016; Eqs. (C4-C5)](@cite) and [haxeltine
 """
 @inline function compute_assimilation_factors(
         photo::LUEPhotosynthesis{NF},
-        constants::PhysicalConstants{NF},
+        constants::MaterialConstants{NF},
```
bgroenks96, line 202: Yeah I think I like this! Much clearer dependency structure.

src/processes/physics_utils.jl
```diff
@@ -30,54 +30,56 @@ Compute partial pressure of CO2 from surface pressure and CO2 concentration in P
...
-@inline vapor_pressure_to_specific_humidity(e, p, ε) = ε * e / p
+@inline function relative_to_specific_humidity(r_h, pr, T, c::ThermodynamicConstants)
+    T_K = celsius_to_kelvin(c, T)
+    phase = T <= zero(T) ? Thermodynamics.Ice() : Thermodynamics.Liquid()
+    return Thermodynamics.q_vap_from_RH(c, pr, T_K, r_h / 100, phase)
+end
 
 """
-    specific_humidity_to_vapor_pressure(q, p, ε)
+    vapor_pressure_to_specific_humidity(c::ThermodynamicConstants, e, pr)
```
bgroenks96, line 47: Can you explain this change? Why does this need to be inverted now?

bgroenks96 (changes requested):
Looks good, great work! Just a few more comments, though, sorry 😅

docs/make.jl
bgroenks96: I think it inserts the given expression in every doctest so you don't need a `@meta` block everywhere.

src/processes/physical_constants.jl
```diff
@@ -1,75 +1,191 @@
...
+```
 
-    "Specific latent heat of sublimation of water in J/kg"
-    Lsg::NF = 2.834e6
+Properties:
+$FIELDS
+"""
+@kwdef struct MaterialConstants{NF}
+    "Density of water in kg/m^3"
+    water_density::NF = 1000.0
```
bgroenks96: I guess this should be `density_water` and `density_ice` if we want to be consistent with the pattern we have started following elsewhere of `[quantity]_[constituent]_[location]`.

docs/src/processes/utils/physical_constants.md
olivierbonte: I don't think there will be situations where you would add constants not add reference I think. The only thing that could exist is a temperature dependent latent heat following https://clima.github.io/Thermodynamics.jl/dev/Formulation/#5.-Latent-Heats (see e.g. https://clima.github.io/Thermodynamics.jl/dev/API/#Thermodynamics.latent_heat_fusion). But maybe that is not a sufficient reason to keep the `_at_reference`? I can remove if it if you prefer

docs/make.jl
olivierbonte: I got it from the Oceananigans documentation in their [make.jl](https://github.com/CliMA/Oceananigans.jl/blob/15b42a614d5b9afa8044c7f3fd51a385aa40b8a1/docs/make.jl#L234)! Following the Documenter.jl documentation [here](https://documenter.juliadocs.org/stable/lib/public/#Documenter.DocMeta.setdocmeta!) and [here](https://documenter.juliadocs.org/stable/man/doctests/#DocTestSetup-and-DocTestTeardown-in-@meta-blocks), I thinki it allows to, before every jldoctest bloc,  add a `using Terrarium` in a `@meta` block as Brian says.

src/processes/physical_constants.jl
olivierbonte: I like those names! I'll adapt

src/processes/physical_constants.jl
olivierbonte: good point, forgot to change this one around to the more spelled out style

src/processes/surface_energy/turbulent_fluxes.jl
olivierbonte: I'll remove the `let` blocks then

src/processes/physical_constants.jl
olivierbonte: Agree that this is better for consitency!

src/processes/physics_utils.jl
olivierbonte: The rational is explained in the documentation at `physics_utils.md`: q -> e is already present in Thermodynamics.jl (https://github.com/NumericalEarth/Terrarium.jl/pull/112/changes#diff-d574d5258fa896b9f1943282cb6128448601c2c88eb245ea0d61734a43cb944dR28), while e -> q is not so therefore this function is added

docs/src/processes/utils/physical_constants.md
olivierbonte: I'll just remove, I suppose it is evident that the constant should be at the reference temperature. If the latent heat at another temperature is needed, this can always be added to the docs for clarity.

olivierbonte: Resolved in https://github.com/NumericalEarth/Terrarium.jl/pull/112/commits/76d4425b9f333cd2914f250b3738706bdc20c103

src/processes/physical_constants.jl
olivierbonte: resolved in https://github.com/NumericalEarth/Terrarium.jl/pull/112/commits/6c38ab7bc5de4c800933a13af75c604d75c011b6

src/processes/surface_energy/turbulent_fluxes.jl
olivierbonte: resolved in https://github.com/NumericalEarth/Terrarium.jl/pull/112/commits/8ca531685f3f0b2f431d701b1692396a8f92656b

src/processes/surface_energy/turbulent_fluxes.jl
olivierbonte: resolved in https://github.com/NumericalEarth/Terrarium.jl/pull/112/commits/8ca531685f3f0b2f431d701b1692396a8f92656b

src/processes/surface_energy/turbulent_fluxes.jl
olivierbonte: resolved in https://github.com/NumericalEarth/Terrarium.jl/pull/112/commits/8ca531685f3f0b2f431d701b1692396a8f92656b

src/processes/physical_constants.jl
olivierbonte: resolved in https://github.com/NumericalEarth/Terrarium.jl/pull/112/commits/aaeccf6914526fc74238bd154f49cc902e76923e

src/processes/physical_constants.jl
olivierbonte: Resolved in https://github.com/NumericalEarth/Terrarium.jl/pull/112/commits/cd6788455c9bbb384b4370ea09cc55940e752286

olivierbonte:
I think everything is addressed now @bgroenks96! If anything is still uncelar let me know and otherwise you can merge once the checks complete :)

src/processes/atmosphere/prescribed_atmosphere.jl
```diff
@@ -1,3 +1,5 @@
+import Thermodynamics
```
bgroenks96: I don't think this should be needed if the `import` is already present in `Terrarium.jl`.

src/processes/physical_constants.jl
```diff
@@ -1,75 +1,191 @@
...
+@inline stefan_boltzmann(c::UniversalConstants, T, ϵ) = ϵ * c.stefan_boltzmann_constant * T^4
 
 """
-    psychrometric_constant(c::PhysicalConstants, p)
+    psychrometric_constant(c::ThermodynamicConstants, p)
 
 Calcualte the psychrometric constant at the given atmospheric pressure `p`.
 """
-@inline psychrometric_constant(c::PhysicalConstants, p) = c.cₐ * p / (c.Llg * c.ε)
+@inline psychrometric_constant(c::ThermodynamicConstants, p) = c.specific_heat_capacity_dry_air * p / (c.latent_heat_vaporization * ε(c))
```
bgroenks96: Shouldn't the epsilon here be `ratio_gas_constant_dry_air_to_water_vapor`?

bgroenks96 (changes requested):

olivierbonte:
Sorry for missing those last things, should be fixed now

docs/make.jl
maximilian-gelbrecht: Cool, yeah that's practical.

maximilian-gelbrecht:
Would you mind still applying the code formatting with [Runic](https://github.com/fredrikekre/Runic.jl)? Then this can be probably merged when @bgroenks96 agrees.

bgroenks96:
Should be fixed in [c8b39a2](https://github.com/NumericalEarth/Terrarium.jl/pull/112/commits/c8b39a242897bd1f95f62828f985d834f9d3e90a)

bgroenks96 (approved):

bgroenks96:
Thanks a lot @olivierbonte for your work on this!

PR #119, opened 2026-05-13
Version bump 0.1.1

maximilian-gelbrecht:
There are a lot of compat and dependency issues with #118 and other in SpeedyWeather.jl, originally caused by the missing compat fixed in #117. I hope just releasing a new minor version fixes this.


bgroenks96:
I would like to merge #112 before releasting 0.1.1.

bgroenks96:
I also don't understand why you need a new version released to fix this? You shouldn't be using the released version in a PR branch anyway. You should be using `[sources]` or a `dev` version?

maximilian-gelbrecht:
> I also don't understand why you need a new version released to fix this? You shouldn't be using the released version in a PR branch anyway. You should be using `[sources]` or a `dev` version?

Across different packages? I disagree with that. Because we wouldn't want to merge https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/1090 with a dependence on an unreleased Terrarium version (also doesn't work in Julia 1.10)

We are not in a major hurry though, so I am also fine holding this of until #112 is finished. Because https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/1090 will probably take some more time anyway. 

Btw: I didn't know that, but it seems you can't put a `[sources]` on a `[weakdeps]` from an extension. Strange.

bgroenks96:
> Across different packages? I disagree with that. Because we wouldn't want to merge https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/1090 with a dependence on an unreleased Terrarium version (also doesn't work in Julia 1.10)

I meant for development, i.e. "in a PR branch".

bgroenks96:
@maximilian-gelbrecht #112 is actually technically a breaking change... so I suppose if we actually follow semantic versioning we should already jump to 0.2. But I think I would rather not... what about you?

maximilian-gelbrecht:
What even means "breaking" for us with Terrarium? Because #112 doesn't change anything in the `examples` so the most user facing functions are not changed. One could say it's not a breaking change then, or not? But we also see Terrarium as a modelling framework. But if that's true basically everything we change that isn't just adding additional processes could qualify as a breaking change. 

To be honest, I would also be pragmatic about this. Right now we are in the beginning, it doesn't really matter. Once there's actually a bit of code and projects that actively depend on Terrarium, we'll also have a good understanding of what's breaking and what isn't and then we should really be clear about semantic versioning and follow it relatively strictly. 

Long story short: I am fine with 0.1.1

bgroenks96 (approved):

PR #122, opened 2026-05-27
Split timestepper and cache and revise Heun

maximilian-gelbrecht:
So far the timestepper is still is in the `ModelIntegrator`. If we move it to the model itself, should both `LandModel` and `SoilModel`/`VegetationModel` have it? In this case both should probably by default just point to the same timestepper, but also allow for different ones. What do you think @bgroenks96 

While doing this here, I also revised the Heun solver to not just deep copy the whole `StateVariables` anymore. Now it is a bit smarter about this. 

Next step still is to put the timestepper in the models.

### DIFF
--- a/src/models/coupled/land_model.jl
+++ b/src/models/coupled/land_model.jl
@@ -46,6 +46,7 @@ end
 function initialize(
         model::LandModel{NF};
         clock = Clock(time = zero(NF)),
+        timestepper = nothing,
         boundary_conditions = (;),
         fields = (;),
         input_variables = ()
@@ -62,7 +63,7 @@ function initialize(
     bcs = merge_boundary_conditions(boundary_conditions, ground_heat_flux_bc, infiltration_bc)
     # Merge user-defined fields with BC fields
     fields = merge((; ground_heat_flux, infiltration), fields)
-    return initialize(vars, grid; clock, boundary_conditions = bcs, fields)
+    return initialize(vars, grid; clock, timestepper, boundary_conditions = bcs, fields)
 end
 
 function initialize!(state, model::LandModel)
--- a/src/models/coupled/land_model.jl
+++ b/src/models/coupled/land_model.jl
@@ -46,7 +46,7 @@ end
 function initialize(
         model::LandModel{NF};
         clock = Clock(time = zero(NF)),
-        timestepper = nothing,
+        timestepper = ForwardEuler{NF}(),
         boundary_conditions = (;),
         fields = (;),
         input_variables = ()
--- a/src/state_variables.jl
+++ b/src/state_variables.jl
@@ -25,7 +25,7 @@ struct StateVariables{
     auxiliary::NamedTuple{auxnames, AuxFields}
     inputs::NamedTuple{inputnames, InputFields}
     namespaces::NamedTuple{nsnames, Namespaces}
-    cache::NamedTuple{cachenames, CacheFields}
+    timestepper_cache::NamedTuple{cachenames, CacheFields}
     clock::ClockType
 
     function StateVariables(
@@ -36,7 +36,7 @@ struct StateVariables{
             auxiliary::NamedTuple{auxnames, AuxFields},
             inputs::NamedTuple{inputnames, InputFields},
             namespaces::NamedTuple{nsnames, Namespaces},
-            cache::NamedTuple{cachenames, CacheFields},
+            timestepper_cache::NamedTuple{cachenames, CacheFields},
             clock::ClockType,
         ) where {
             NF, prognames, auxnames, inputnames, nsnames, cachenames,
@@ -51,7 +51,7 @@ struct StateVariables{
             auxiliary,
             inputs,
             namespaces,
-            cache,
+            timestepper_cache,
             clock,
         )
     end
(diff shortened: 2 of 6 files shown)


src/models/coupled/land_model.jl
```diff
@@ -46,6 +46,7 @@ end
 function initialize(
         model::LandModel{NF};
         clock = Clock(time = zero(NF)),
+        timestepper = ForwardEuler{NF}(),
```
maximilian-gelbrecht, line 49: Won't be needed anymore once the timestepper moves inside of the model, but it's here so that the current state of the branch already works.

src/state_variables.jl
```diff
@@ -15,15 +15,17 @@ by the timestepping scheme.
...
         ProgFields, TendFields, AuxFields, InputFields, Namespaces,
+        CacheFields,
         ClockType,
     } <: AbstractStateVariables
     prognostic::NamedTuple{prognames, ProgFields}
     tendencies::NamedTuple{prognames, TendFields}
     auxiliary::NamedTuple{auxnames, AuxFields}
     inputs::NamedTuple{inputnames, InputFields}
     namespaces::NamedTuple{nsnames, Namespaces}
+    cache::NamedTuple{cachenames, CacheFields}
```
bgroenks96: Is this only for the timestepper? It could also be called `timestepper` then or `timestepper_cache`..? Just so it's clear that this is not intended to be used as a general cache.

src/timesteppers/heun.jl
```diff
@@ -3,63 +3,84 @@
...
+    prognostic = map(deepcopy, state.prognostic)
+    tendencies = map(deepcopy, state.tendencies)
+    return (; prognostic, tendencies)
+end
 
-function initialize(timestepper::Heun{NF}, ::AbstractModel{NF2}, state) where {NF, NF2}
-    @assert NF == NF2 "Time stepper was initialized with eltype(Δt)=$(NF) but model has eltype=$(NF2)"
-    return setproperties(timestepper, stage = deepcopy(state))
+# Save current prognostic and tendencies into the Heun cache. Recurses into namespaces.
+function save_heun_cache!(state::StateVariables)
```
bgroenks96: This feels like a Claude-generated name 😅 . The method seems quite general so I don't see why it should be named `save_heun_cache`..?

src/timesteppers/heun.jl
```diff
@@ -3,63 +3,84 @@
...
 end
 
-function average_tendencies!(
-        state::StateVariables{NF, prognames},
-        stage::StateVariables{NF, prognames}
-    ) where {NF, prognames}
-    for name in prognames
-        state.tendencies[name] .= (state.tendencies[name] + stage.tendencies[name]) / 2
+# Restore prognostic fields from the cache. Recurses into namespaces.
+function restore_heun_prognostic!(state::StateVariables)
```
bgroenks96: Same here, can this be generalized?

bgroenks96 (changes requested):

src/timesteppers/heun.jl
maximilian-gelbrecht: For saving the cache I am not so sure. For RK solvers, the cache would be the different steps $$k_i$$ and you'd save those and not just the current prognostic and tendencies, or am I mistaking? 

The specific name choice is also because it doesn't actually dispatch on the timestepper. I guess we could also just add the time stepper as an input and go for the generic name, even if the timestepper isn't used. That's probably a more Julia-like solution.

src/state_variables.jl
maximilian-gelbrecht: I think it would prefer `timestepper_cache` in that case.

src/timesteppers/heun.jl
maximilian-gelbrecht: Yeah, same as above kind of. I do think this is relatively specific to Heun, but nonetheless the more Julia way to do this would be just to dispatch on the timestepper nonetheless and have more generic name.

src/timesteppers/heun.jl
maximilian-gelbrecht: Ok, done that now.

src/state_variables.jl
maximilian-gelbrecht: Done that now.

maximilian-gelbrecht:
I figure this is really quite self contained. We could just merge this and then in a follow up move the timestepper from `ModelIntegrator` into `Model`.

src/timesteppers/heun.jl
maximilian-gelbrecht: I would suggest that we keep it now with the generic name but dispatch on Heun, and then when we implement e.g. an RK3 we can see if we can actually generalise it or not. What do you think?

src/timesteppers/heun.jl
```diff
@@ -3,63 +3,84 @@
...
-    # Update clock
-    tick!(stage.clock, Δt)
-    # Recompute tendencies after timestep
-    update_state!(stage, model, inputs, compute_tendencies = true)
+    closure!(state, model)
+
+    # Recompute tendencies ∂u∂t₁ at the predictor state (u_pred, t + Δt)
+    update_state!(state, model, inputs, compute_tendencies = true)
+    # Average tendencies in place: state.tendencies ← (∂u∂t₀ + ∂u∂t₁) / 2
+    average_tendencies!(timestepper, state)
```
bgroenks96: The argument order should be `state, timestepper` to be consistent with Julia conventions.

src/state_variables.jl
```diff
@@ -364,7 +374,20 @@ function initialize(
     end
     # get closure variable names
     closurenames = map(varname, closure_variables(values(vars.prognostic)))
-    # construct and return StateVariables
+    # construct StateVariables with an empty cache; the timestepper-specific cache
+    # is allocated below now that all other state variables have been initialized
+    state = StateVariables(
```
bgroenks96: very small thing but I would give this a different name like `initial_state`

bgroenks96 (changes requested):

src/state_variables.jl
maximilian-gelbrecht: done

src/timesteppers/heun.jl
maximilian-gelbrecht: done

bgroenks96 (approved):

PR #125, opened 2026-06-01
 Timestepper revision: timesteppers as part of the models and allow for multiple ones

maximilian-gelbrecht:
Major breaking change that makes the timesteppers part of each `AbstractModel` and allows for multiple timesteppers to be defined (at the moment precisely two: `explicit` and `implicit`, but in principle we could loosen this in the future if needed). Each prognostic variable is defined with an `explicit` or `implicit` keyword to define which time stepper should be used on it. This keyword becomes a type parameter, to make the whole approach fully type stable. 

To accommodate for this I also had to introduce another major breaking change: `state = initialize(model, args..)` was renamed to `state = StateVariables(model, args...)` to avoid the ambiguity with `model_integrator = initialize(model, args...)` since timesteppers are no longer used in this function call. 

The unit tests already pass locally, so the basics of this work. But of course they only work with the standard explicit solvers so far.  It's not exactly clear yet to me whether we need further modifications for the implicit solver, especially with how to set up the equation system for the implicit solver of our choice. 

The Speedy coupling example is expected to fail as the `SpeedyWeatherTerrariumExt` needs to be revised as well (`initialize` -> `StateVariables`). 

Further adjustments to the docs would be needed to explain those changes there as well.


src/timesteppers/model_integrator.jl
```diff
@@ -121,40 +139,68 @@ Advance the model forward by one timestep with optional timestep size `Δt`. If
...
 
+"""
+    timestep!(integrator::ModelIntegrator, timesteppers::NamedTuple, Δt)
+
+Advance the model forward by one timestep of size `Δt`, integrating each prognostic variable with the
+time stepper it was assigned via its `timestepper` (`:explicit` or `:implicit`). The prognostic variables
+are grouped by their assigned time stepper, and each group is forwarded to the corresponding
+`timestep!(integrator, timestepper, Δt, names)` method. The clock is advanced once for the whole step.
+"""
+function timestep!(integrator::ModelIntegrator, timesteppers::NamedTuple, Δt)
```
maximilian-gelbrecht: This is the main orchestrator of the timestepping. At the moment that's of course super naive approach. But I think every concrete version also depends on what we end up doing in the end with the implicit timesteppers. 

We may also just forward this to  `timestep!(::Model, ::Timestepper,  Δt)` routines for dispatching

docs/src/models/land_model.md
```diff
@@ -18,8 +18,8 @@ using Terrarium
 ```@example landmodel
 arch = CPU()
 grid = ColumnGrid(arch, Float32, ExponentialSpacing(N = 10)) # 10 soil layers
-model = LandModel(grid) # Default configuration
-integrator = initialize(model, ForwardEuler(eltype(grid)))
+model = LandModel(grid, timesteppers=ForwardEuler(eltype(grid))) # Default configuration
```
bgroenks96: I would prefer to just have it be `timestepper` and then have composite timestepper types that allow for combining multiple timesteppers together, e.g. `IMEXTimestepper(implicit, explicit)`.

examples/extending/linear_heat_conduction.jl
```diff
@@ -172,18 +172,20 @@ end
...
         NF,
         Grid <: Terrarium.AbstractLandGrid{NF},
         Cond <: AbstractHeatConduction{NF},
         Init <: Terrarium.AbstractInitializer{NF},
+        TS <: NamedTuple,
     } <: Terrarium.AbstractModel{NF, Grid}
     grid::Grid
     conduction::Cond = LinearHeatConduction(eltype(grid))
     initializer::Init = DefaultInitializer(eltype(grid))
+    timesteppers::TS = (; explicit = ForwardEuler(eltype(grid)))
```
bgroenks96: See again my comment above. No `NamedTuple`, just a single composite timestepper.

examples/simulations/soil_heat_column.jl
```diff
@@ -22,18 +22,18 @@ initializer = SoilInitializer(
     hydrology = ConstantSaturation(eltype(grid), sat = 1.0)
 )
 
-# We're now ready to create our model:
-model = SoilModel(grid; initializer)
+# We're now ready to create our model.
+# We also choose the time stepper to be [`ForwardEuler`](@ref) for this example.
```
bgroenks96, line 26: Is it not the default?

src/timesteppers/abstract_timestepper.jl
```diff
@@ -10,20 +10,107 @@ Base type for time steppers.
...
+"""
+abstract type AbstractImplicitTimestepper{NF} <: AbstractTimeStepper{NF} end
+
+"""
+    $SIGNATURES
+
+Return the class (`:explicit` or `:implicit`) that the given `timestepper` fills within a model's
+`timesteppers` `NamedTuple`. This is determined by the abstract supertype of the time stepper.
+"""
+timestepper_class(::AbstractExplicitTimestepper) = :explicit
```
bgroenks96: Yeah not a huge fan of this... and I think the need for it will go away when you just use a composite timestepper type, although we could still create a proper trait type if needed.

src/timesteppers/abstract_timestepper.jl
```diff
@@ -10,20 +10,107 @@ Base type for time steppers.
...
+
+# TODO: currently at most one per class of timestepper, we might relax that in the future
+"""
+    to_timesteppers(::Type{NF}, timesteppers)
+
+Normalize a single timestepper, a tuple/vector of timesteppers, or a `NamedTuple` into the canonical
+`timesteppers` `NamedTuple` with (at most) the entries `explicit` and `implicit`. If no explicit timestepper 
+is provided, the `explicit` entry defaults to [`ForwardEuler`](@ref).
+"""
+to_timesteppers(::Type{NF}, timestepper::AbstractTimeStepper) where {NF} = to_timesteppers(NF, (timestepper,))
```
bgroenks96: Unless I am missing something, I think this could also be removed if we just have `IMEXTimestepper`.

src/timesteppers/abstract_timestepper.jl
bgroenks96: I also think a trait would be anyway better than relying on inheritance.

src/timesteppers/model_integrator.jl
bgroenks96: Agreed. We could also consider keeping such dispatches in the timestepping code, although I forget at the moment what the import order is... might not work actually.

src/timesteppers/model_integrator.jl
```diff
@@ -121,40 +139,68 @@ Advance the model forward by one timestep with optional timestep size `Δt`. If
 `compute_auxiliary!` is called after the time step in order to update the values of auxiliary/diagnostic
 variables.
 """
-timestep!(integrator::ModelIntegrator; finalize = true) = timestep!(integrator, default_dt(timestepper(integrator)); finalize)
+timestep!(integrator::ModelIntegrator; finalize = true) = timestep!(integrator, default_dt(get_timesteppers(integrator.model)); finalize)
```
bgroenks96: Should also be changed to `get_timestepper`

src/abstract_model.jl
```diff
@@ -215,12 +215,20 @@ defined by the coupling interface for the process type.
...
-(::Type{Model})(grid::AbstractLandGrid, args...; kwargs...) where {Model <: AbstractModel} = Model(args...; grid, kwargs...)
+as the first positional argument. The `timesteppers` keyword can be a single timestepper, 
+a tuple/vector of timesteppers, or a `NamedTuple`.  All converted into the canonical 
+`(; explicit[, implicit])` form.
+"""
+function (::Type{Model})(grid::AbstractLandGrid, args...; timesteppers = nothing, kwargs...) where {Model <: AbstractModel}
+    return if isnothing(timesteppers)
+        Model(args...; grid, kwargs...)
+    else
+        Model(args...; grid, timesteppers = to_timesteppers(eltype(grid), timesteppers), kwargs...)
```
bgroenks96: I'm not sure I understand why this is needed? Is this same logic also in the keyword arg constructors in the types themselves?

src/state_variables.jl
```diff
@@ -31,6 +31,7 @@ struct StateVariables{
     function StateVariables(
             ::Type{NF},
             closurenames::Tuple{Vararg{Symbol}},
+            timestepper_classes::Tuple{Vararg{Symbol}},
```
bgroenks96: As discussed previously, I would prefer to not store "timestepper classes" in the `StateVariables` type and rather store any such relevant information in the `cache`. The cache can also itself be a `struct` on which we can dispatch.

src/state_variables.jl
```diff
@@ -370,15 +380,22 @@ function initialize(
...
+        ns_classes = get(timestepper_classes, varname(ns), (;))
+        StateVariables(ns.vars, grid; clock, timestepper_classes = ns_classes, boundary_conditions = ns_bcs, fields = ns_fields)
     end
     # get closure variable names
     closurenames = map(varname, closure_variables(values(vars.prognostic)))
+    # Resolve the timestepper class (:explicit/:implicit) of each prognostic variable, in the same order as
+    # the names. The default class is the one each variable was declared with (see `prognostic`); the
+    # `timestepper_classes` keyword overrides it per-variable. The resolved classes are stored as a type
+    # parameter so that `prognostic_names(state, class)` can select per class in a type-stable way.
+    resolved_classes = consolidate_timestepper_classes(vars.prognostic, vars.namespaces, timestepper_classes)
```
bgroenks96: Hopefully the need for this will go away with the explicit composite timestepper type? Not sure...

bgroenks96 (changes requested):

bgroenks96:
As discussed, let's change the `NamedTuple` of timesteppers pattern to just be a single `timestepper` that can wrap two or more other timesteppers. Models are free to place constraints on which timestepper types they support.

examples/simulations/soil_heat_column.jl
maximilian-gelbrecht: It is the default, but we called it explicitly in this example before (see below)

src/state_variables.jl
maximilian-gelbrecht: Yeah, the only way to do it would be actually turn it into a `struct` then, because we need it as type parameters for type stability, and not just has an element of the tuple. 

I guess that would work, yes.

src/state_variables.jl
maximilian-gelbrecht: No not really. This stayed for the moment. 

Do you disagree with the general logic of having defaults that can then be overwritten at model construction?

examples/extending/linear_heat_conduction.jl
maximilian-gelbrecht: Done now

src/timesteppers/abstract_timestepper.jl
```diff
@@ -69,9 +126,6 @@ function explicit_step!(state, grid::AbstractLandGrid, timestepper::AbstractTime
         # debug site post-step
         debugsite!(explicit_step!, state.prognostic[name], name)
     end
-    fastiterate(state.namespaces) do ns
```
maximilian-gelbrecht, line 72: To be honest I just find the `namespaces` confusing for implementations. We don't use them at the moment, and maybe it might also just be okay to remove them completely (in another PR)

src/timesteppers/imex.jl
```diff
@@ -0,0 +1,144 @@
...
+variable is routed to the explicit or implicit sub-stepper according to the resolved classes stored in the
+[`IMEXCache`](@ref) type; each sub-stepper fetches its own sub-cache via [`get_cache`](@ref). The clock is
+advanced once for the whole step.
+"""
+function timestep!(integrator::ModelIntegrator, timestepper::IMEX, Δt)
+    cache = integrator.state.timestepper_cache
+    # type-stable split of the prognostic variables by their resolved class
+    explicit_names = prognostic_names(integrator.state, cache, Val(:explicit))
+    implicit_names = prognostic_names(integrator.state, cache, Val(:implicit))
+    isempty(explicit_names) || timestep!(integrator, timestepper.explicit, Δt, explicit_names)
```
maximilian-gelbrecht: Currently it's explicit first, then implicit, but we'll revisit this once we implement solvers. Clearly one could also dispatch on the types within the `IMEX` here, to have different orders for different pairs.

src/timesteppers/abstract_timestepper.jl
maximilian-gelbrecht: yes, removed now

src/timesteppers/model_integrator.jl
maximilian-gelbrecht: done

src/abstract_model.jl
maximilian-gelbrecht: not needed anymore with the `IMEX` type, removed

src/state_variables.jl
maximilian-gelbrecht: Okay, done with a `struct`. You were right, that's better.

src/timesteppers/abstract_timestepper.jl
bgroenks96: We will need them for tiling and PFTs. Unless you have a suggestion for an alternative way to handle that, they have to stay.

src/timesteppers/imex.jl
bgroenks96: isn't IMEX usually explicit first and then implicit? the implicit solver needs the tendencies from the explicit step on the RHS, no?

src/timesteppers/model_integrator.jl
```diff
@@ -31,9 +31,25 @@ struct ModelIntegrator{
 
     "Optional named tuple of user-specified field initializers"
     initializers::Inits
+end
 
-    "Time stepper"
-    timestepper::TimeStepper
+# Outer constructor so that `ModelIntegrator` can be constructed with a timestepper type
+# so that it can correctly subtype Oceananigans.AbstractModel
+# TODO: Is this actually needed?
```
bgroenks96: with the current Oceananigans API it is needed. The `timestepper` type should be `IMEX`, not the explicit one.

src/abstract_variables.jl
```diff
@@ -191,9 +191,16 @@ point in (simulation) time. From a computational perspective, they can be seen a
...
+
+Each prognostic variable is assigned a `timestepper` class (`:explicit` or `:implicit`) which, when the
+model's `timestepper` is an [`IMEX`](@ref), determines whether it is integrated by the `explicit` or
+`implicit` timestepper. A single (non-`IMEX`) timestepper integrates all prognostic variables regardless
+of class. The class declared here is a default that can be overridden per-variable via the
+`timestepper_classes` argument of the [`IMEX`](@ref) constructor.
 """
 struct PrognosticVariable{
         name,
+        timestepper,
```
bgroenks96: I still think I would prefer to have methods `prognostic_names(::AbstractModel, ::AbstractTimestepper, ::Val{:explicit})` (maybe make a trait type `Implicit`/`Explicit` instead of `Val`) rather than directly put this in the `PrognosticVariable` type. This would then allow the variables to be explicitly configured for different timestepper configurations.

bgroenks96 (changes requested):

src/timesteppers/imex.jl
```diff
@@ -0,0 +1,148 @@
...
+        TC <: NamedTuple,
+    } <: AbstractTimeStepper{NF}
+    "Sub-stepper for prognostic variables of class `:explicit`"
+    explicit::E
+
+    "Sub-stepper for prognostic variables of class `:implicit`"
+    implicit::I
+
+    "Per-variable timestepper class overrides (`varname => :explicit`/`:implicit`)"
+    timestepper_classes::TC
```
bgroenks96: As discussed, I would prefer to remove this.

src/state_variables.jl
bgroenks96: As discussed, I think the timestepper "classes" should declared by a method `prognostic_names` (or something like that) and then collected when constructing the cache. They do not need to be in the `IMEX` type itself. Then you can eliminate this added complexity of merging them, I think.

src/timesteppers/abstract_timestepper.jl
bgroenks96: and they need to be handled here, not just deleted.

src/timesteppers/model_integrator.jl
maximilian-gelbrecht: The comment is outdated, it takes the IMEX

src/timesteppers/abstract_timestepper.jl
maximilian-gelbrecht: reintroduced as part of the revised approach

src/abstract_variables.jl
maximilian-gelbrecht: Outdated

src/timesteppers/imex.jl
maximilian-gelbrecht: Removed / outdated

src/state_variables.jl
maximilian-gelbrecht: Removed / outdated

src/timesteppers/abstract_timestepper.jl
maximilian-gelbrecht: Removed / outdated

src/timesteppers/imex.jl
maximilian-gelbrecht: Yes, probably. That's also still like this even in the revised version

maximilian-gelbrecht:
If you just want to see the revised workflow, probably best to look at the IMEX unit tests. 

One thing to mention, currently this approach with 
```Julia 
    Terrarium.timestepping(::AbstractVariable{:b}, ::TwoVarModel, ::AbstractIMEX) = Implicit()
```

Is using globale variable name definitions, so the line above would declare `b` in all namespaces as implicit. This might be desired. Maybe we also don't like this, then we would have to use variable paths as in the namespaces + input PR and change something about how the time steppers recurse over namespaces.

docs/src/extending/state_variables.md
```diff
@@ -20,7 +20,7 @@ InputVariable
 ```
 Variables are typically constructed using one of the convenience functions [`prognostic`](@ref), [`auxiliary`](@ref), or [`input`](@ref).
 
-These variable definitions are purely symbolic; they do not hold any data and cannot be used for computation. Calling [`initialize`](@ref) on a model, process, or [`Variables`](@ref) container (see following sections) results in corresponding [Fields](@ref) being allocated for each variable. 
+These variable definitions are purely symbolic; they do not hold any data and cannot be used for computation. Calling [`StateVariables`](@ref) on a model, process, or [`Variables`](@ref) container (see following sections) results in corresponding [Fields](@ref) being allocated for each variable.
```
bgroenks96:
```suggestion
These variable definitions are purely symbolic; they do not hold any data and cannot be used for computation. Constructing [`StateVariables`](@ref) from a model, process, or [`Variables`](@ref) container (see following sections) results in corresponding [Fields](@ref) being allocated for each variable. 
```

docs/src/models/land_model.md
```diff
@@ -18,8 +18,8 @@ using Terrarium
 ```@example landmodel
 arch = CPU()
 grid = ColumnGrid(arch, Float32, ExponentialSpacing(N = 10)) # 10 soil layers
-model = LandModel(grid) # Default configuration
-integrator = initialize(model, ForwardEuler(eltype(grid)))
+model = LandModel(grid, timestepper=ForwardEuler(eltype(grid))) # Default configuration
```
bgroenks96: I get that this is a direct translation of the existing example, but I would prefer to just remove the `timestepper=ForwardEuler(eltype(grid))` in all cases where this already matches the default. The only exception to this would be for examples that are specifically trying to demonstrate something about the timestepping.

bgroenks96:
```suggestion
model = LandModel(grid) # Default configuration
```

docs/src/models/soil_model.md
```diff
@@ -18,8 +18,8 @@ using Terrarium
 ```@example soilmodel
 arch = CPU()
 grid = ColumnGrid(arch, Float32, ExponentialSpacing(N = 10)) # 10 soil layers
-model = SoilModel(grid)
-integrator = initialize(model, ForwardEuler(eltype(grid)))
+model = SoilModel(grid, timestepper=ForwardEuler(eltype(grid)))
```
bgroenks96:
```suggestion
model = SoilModel(grid)
```

docs/src/models/vegetation_model.md
```diff
@@ -18,8 +18,8 @@ using Terrarium
 ```@example vegmodel
 arch = CPU()
 grid = ColumnGrid(arch, Float32, UniformSpacing(N = 1)) # grid with one vertical layer
-model = VegetationModel(grid) # Default configuration
-integrator = initialize(model, ForwardEuler(eltype(grid)))
+model = VegetationModel(grid, timestepper=ForwardEuler(eltype(grid))) # Default configuration
```
bgroenks96:
```suggestion
model = VegetationModel(grid) # Default configuration
```

docs/src/running/initialization.md
```diff
@@ -6,11 +6,11 @@ CurrentModule = Terrarium
 
 ## Overview
 
-Model and process types in Terrarium are *stateless* and *immutable*; i.e. they only specify parameters and model configuration. To allocate state variables for a model or process, we need to use the [`initialize`](@ref) method:
+Model and process types in Terrarium are *stateless* and *immutable*; i.e. they only specify parameters and model configuration. To allocate state variables for a model or process, we use the [`StateVariables`](@ref) constructor:
```
bgroenks96, line 9: While I do understand the need for this change, I have to say that I kind of miss the simplicity of just using `initialize` for this. Is there any way we could change the signature to resolve the ambiguity?

bgroenks96: at the very least we should probably consider (in another PR) to finally rename `StateVariables`. It's a bad name considering that we try to distinguish in Terrarium between (symbolic) "variables" and `Field`s that hold actual data.

docs/src/index.md
```diff
@@ -59,10 +59,10 @@ using Terrarium
 num_columns = 1
 arch = CPU()
 grid = ColumnGrid(arch, Float32, ExponentialSpacing(N=10), num_columns)
-model = SoilModel(grid)
+model = SoilModel(grid, timestepper=ForwardEuler(eltype(grid)))
```
bgroenks96:
```suggestion
model = SoilModel(grid)
```

docs/src/index.md
```diff
@@ -81,10 +81,10 @@ using CUDA # needs to be separately installed
 rings = FullGaussianGrid(8) # Gaussian grid with 16 latitudinal rings, 8 per hemisphere (512 points, ~9.0˚ lat/lon)
 arch = CUDA.functional() ? GPU() : CPU() # run on the GPU (if available)
 grid = ColumnRingGrid(arch, Float32, ExponentialSpacing(N=10), rings) # create grid
-model = SoilModel(grid)
+model = SoilModel(grid, timestepper=ForwardEuler(eltype(grid)))
```
bgroenks96:
```suggestion
model = SoilModel(grid)
```

src/timesteppers/abstract_timestepper.jl
```diff
@@ -1,15 +1,66 @@
...
 Base type for time steppers.
 """
 abstract type AbstractTimeStepper{NF} end
 
+"""
+    $TYPEDEF
+
+Base type for *explicit* time steppers.
+"""
+abstract type AbstractExplicitTimestepper{NF} <: AbstractTimeStepper{NF} end
```
bgroenks96: I know I raised this point before, but I think it might be better to make this a trait rather than rely on subtyping? Or do you have a specific reason for preferring subtyping here?

src/timesteppers/abstract_timestepper.jl
```diff
@@ -1,15 +1,66 @@
...
+    $TYPEDEF
+
+Base type for *implicit* time steppers.
+"""
+abstract type AbstractImplicitTimestepper{NF} <: AbstractTimeStepper{NF} end
+
+"""
+    $TYPEDEF
+
+Indicator type marking a prognostic variable to be integrated by the *explicit* sub-stepper of an
```
bgroenks96: You could have both of these subtype an abstract type and then call it a trait.

src/timesteppers/abstract_timestepper.jl
```diff
@@ -1,15 +1,66 @@
...
+
+"""
+    timestepping(var::AbstractVariable, model::AbstractModel, timestepper::AbstractTimeStepper)
+
+Return the timestepping class — [`Explicit`](@ref) or [`Implicit`](@ref) — with which the prognostic
+variable `var` of `model` is integrated under `timestepper`. The default is `Explicit()` for all variables,
+models, and timesteppers. Specialize this method (typically on an [`AbstractIMEX`](@ref) timestepper together
+with particular variable and/or model types) to route selected variables to the implicit sub-stepper.
+"""
+timestepping(::AbstractVariable, model, ::AbstractTimeStepper) = Explicit()
```
bgroenks96, line 64: and we could reuse the same trait both for variables and timesteppers themselves, since both can be associated with a particular type of timestepping.

src/timesteppers/abstract_timestepper.jl
```diff
@@ -24,6 +75,31 @@ Return `true` if the given time stepper is adaptive, false otherwise.
 """
 function is_adaptive end
 
+"""
+    $SIGNATURES
+
+Default `timestepper` for models: a single `explicit` [`ForwardEuler`](@ref) time stepper.
+"""
+default_timestepper(::Type{NF}) where {NF} = ForwardEuler(NF)
```
bgroenks96, line 85: Hmm... shouldn't the default timestepper be model-dependent? Why do we need this? There's also nothing for it to meaningfully dispatch on.

src/timesteppers/forward_euler.jl
```diff
@@ -14,16 +14,15 @@ default_dt(euler::ForwardEuler) = euler.Δt
 
 is_adaptive(euler::ForwardEuler) = false
 
-function timestep!(integrator::ModelIntegrator, timestepper::ForwardEuler, Δt)
+# Step only the prognostic variables named in `names` with the forward Euler scheme.
+function timestep!(integrator::ModelIntegrator, timestepper::ForwardEuler, Δt, names::Tuple)
```
bgroenks96, line 20: I still find this pattern with the `names` argument here a bit strange and I think it also won't be type stable in the later `fastiterate` call...?

src/timesteppers/imex.jl
```diff
@@ -0,0 +1,153 @@
...
+
+Advance the model forward by one timestep of size `Δt` using an [`AbstractIMEX`](@ref) timestepper. Each
+prognostic variable is routed to the explicit or implicit sub-stepper according to the resolved classes stored
+in the [`IMEXCache`](@ref) type; each sub-stepper fetches its own sub-cache via [`get_cache`](@ref). The clock
+is advanced once for the whole step.
+"""
+function timestep!(integrator::ModelIntegrator, timestepper::AbstractIMEX, Δt)
+    cache = integrator.state.timestepper_cache
+    # type-stable split of the prognostic variables by their resolved class
+    explicit_names = prognostic_names(integrator.state, cache, Explicit())
```
bgroenks96, line 140: I guess here is where I don't understand why this needs to be done in `IMEX` rather than directly in the timestepper dispatches for `timestep!`.

src/timesteppers/forward_euler.jl
bgroenks96: ah maybe it will be stable with constant propagation

src/timesteppers/imex.jl
```diff
@@ -0,0 +1,153 @@
...
+
+Construct an [`IMEX`](@ref) time stepper from an `explicit` and an `implicit` sub-stepper (which must share
+the same numerical type `NF`). Which prognostic variables are integrated implicitly is controlled by
+specializing [`timestepping`](@ref).
+"""
+IMEX(; explicit, implicit) = IMEX(explicit, implicit)
+
+default_dt(imex::AbstractIMEX) = default_dt(explicit_timestepper(imex))
+
+is_adaptive(imex::AbstractIMEX) = is_adaptive(explicit_timestepper(imex)) || is_adaptive(implicit_timestepper(imex))
```
bgroenks96, line 67: Do we currently have a way for either of the substeppers to "request" or choose a timestep? I guess not. In Oceananigans, this is actually done externally via a callback... which I am still not so sure I like.

bgroenks96 (changes requested):

docs/src/models/land_model.md
maximilian-gelbrecht: Agreed and changed.

src/state_variables.jl
```diff
@@ -338,20 +334,20 @@ Initialize a `StateVariables` data structure containing `Field`s defined on the
 for all variables defined by `process`. Any predefined `boundary_conditions` and `fields` will
 be passed through to `initialize` for each variable.
 """
-function initialize(
+function StateVariables(
```
bgroenks96, line 337: Why was this necessary?

src/state_variables.jl
```diff
@@ -362,13 +358,15 @@ end
 
 Initialize a `StateVariables` data structure containing `Field`s defined on the given `grid`
 for all variables in `vars`. Any predefined `boundary_conditions` and `fields` will be passed
-through to `initialize` for each variable.
+through to `initialize` for each variable. The `timestepper`'s cache is allocated via
+`initialize(timestepper, state, progvars)`.
 """
-function initialize(
+function StateVariables(
```
bgroenks96, line 364: also here

docs/src/running/initialization.md
maximilian-gelbrecht: I would argue against this. I don't really like `initialize` or `initialize!` naming for functions that are actually just constructing an instance of a type. In this case the function acts as a constructor, so it should just be that: a constructor of `StateVariables`. 

Renaming `StateVariables` is absolutely fine though. Maybe just `State`? Or what are you thinking about?

src/timesteppers/abstract_timestepper.jl
maximilian-gelbrecht: With trait you mean type parameter?

src/timesteppers/abstract_timestepper.jl
maximilian-gelbrecht: Currently this is just used as central definition of the default timestepper. Each model just has 
```
@component timestepper::Timestepper = default_timestepper(eltype(grid))
```

I do prefer this to hard coding a default one into all models. This way we can change it easily later. 

We could also just have this as a `const` somewhere though, doesn't need to be a function.

src/timesteppers/forward_euler.jl
maximilian-gelbrecht: I think it should be type stable. Otherwise we would also see some problems in the unit tests, or not?

src/timesteppers/imex.jl
maximilian-gelbrecht: No, currently not, but couldn't you do this within a constructor of `IMEX`? So the implicit and explicit timestepper get constructed first, and then, if defined, the other gets reconstructed with the requested timestep? Something like this?

docs/src/running/initialization.md
bgroenks96: Maybe `ModelState` or `StateFields`?

`ModelState` is slightly misleading since it can be constructed for individual processes (not just models) but I guess that's OK.

maximilian-gelbrecht: I think it like `StateFields` more of both of those.

src/timesteppers/abstract_timestepper.jl
bgroenks96: No. A [trait](https://en.wikipedia.org/wiki/Trait_(computer_programming)) is a lightweight, functional assignment of a type to a group or category. For example:

```julia
abstract type Timestepping end
struct Explicit <: Timestepping end
struct Implicit <: Timestepping end

timestepping(::ForwardEuler) = Explicit()
```

Traits are usually associated with a method interface for acting on objects that have that trait. In Julia, you can realize this with a forwarding dispatch:

```julia
timestep!(integrator, timestepper, dt) = timestep!(integrator, timestepper,  timestepping(timestepper), dt)
timestep!(integrator, timestepper, ::Explicit, dt) = ...
timestep!(integrator, timestepper, ::Implicit, dt) = ...
```

Traits are more flexible than subtyping because types can be assigned multiple different traits and don't have to be forced into a single hierarchy.

See Lyndon White's old [blog post](https://www.juliabloggers.com/the-emergent-features-of-julialang-part-ii-traits/) on this as an emergent feature of Julia. In other languages like Rust and Scala, they are first-class features that the compiler can type check.

src/timesteppers/abstract_timestepper.jl
bgroenks96: I am just questioning whether it makes sense to have a single default timestepper for *all* models in the first place. But maybe at least make it dispatch on `grid` rather just `NF`?

src/timesteppers/forward_euler.jl
bgroenks96: I don't remember whether or not we have `@inferred` tests for the timesteppers.

src/timesteppers/imex.jl
bgroenks96: I more meant for adaptive time step sizes.

src/timesteppers/abstract_timestepper.jl
maximilian-gelbrecht: Yes, I understand. But what's the alternative for having one? Just writing `ForwardEuler(NF)` into all models?

src/timesteppers/forward_euler.jl
maximilian-gelbrecht: The Enzyme example would fail in this case as well

src/timesteppers/imex.jl
maximilian-gelbrecht: Ah okay. Couldn't we somehow use the `ParameterEditing` for this? Or is this too hacky?

src/timesteppers/abstract_timestepper.jl
maximilian-gelbrecht: Ok got it, and implemented it

src/timesteppers/abstract_timestepper.jl
maximilian-gelbrecht: Okay, I implemented a trait based version

src/timesteppers/abstract_timestepper.jl
bgroenks96: All models where it makes sense. Probably `LandModel` and `SoilModel` should use `IMEX` by default, once it's working.

docs/src/running/initialization.md
maximilian-gelbrecht: I'll do this change last, before merging, so that the diff remains a bit easier to review

src/timesteppers/imex.jl
bgroenks96: For adapting time step sizes? That would be very hacky. I would also object to characterizing the time step size as a parameter.

src/state_variables.jl
maximilian-gelbrecht: Consistency with the other renaming. I would find odd to keep this as `initialize` when it also just acts as a constructor.

src/state_variables.jl
maximilian-gelbrecht: Same as above.

src/timesteppers/imex.jl
maximilian-gelbrecht: Okay, yes maybe a bit too hacky. 

But in principle, once we have an adaptive time stepper, couldn't we just define the adaptivity also as a trait and then let `IMEX` check first whether one of the time steppers is adaptive and then time step and modify the timestep of the other solver?

bgroenks96: I think IMEX would need to ask both the explicit and implicit timestepper what their maximum permissible timestep is and then take the minimum of the two.

This possibly tricky part is that the answer for each would depend on which variables were assigned to them. We wouldn't want the explicit timestepper reporting a maximum timestep based on the soil prognostics when they are actually being handled by the implicit timestepper.

src/state_variables.jl
bgroenks96: Constructors are just functions, though?

The point of `initialize` was to be an idiom for "take this stateless model structure and allocate all of the necessary state variables". I do agree that it was inconsistent using it both for `StateVariables` and `ModelIntegrator`, but my preferred solution to that would probably be to make `initialize` be *only* for `StateVariables` and then use `integrator` or even just `Simulation` as convenience methods to construct the latter.

src/timesteppers/abstract_timestepper.jl
maximilian-gelbrecht: Then, I'd still have a `DEFAULT_IMEX` and `DEFAULT_EXPLICIT` defined somewhere. This here is just the translation of the `ModelIntegrator` previously also having a default argument for the timestepper (ForwardEuler).

src/state_variables.jl
maximilian-gelbrecht: I am just thinking that when all it does is to construct a type then just using a constructor with the name of the type is the expected way to call it.

src/timesteppers/imex.jl
maximilian-gelbrecht: I would still do this here because it's only actually relevant for IMEX solvers. Only for those we need to split the prognostic variables.

src/timesteppers/imex.jl
maximilian-gelbrecht: I can see that this isn't super easy, but I am confident we can engineer something when it's coming up.

src/timesteppers/abstract_timestepper.jl
bgroenks96: I still don't really see a strong advantage over just having every model declare its own default, but if you prefer having `default_timestepper`, then it's fine.

src/timesteppers/imex.jl
bgroenks96: Ok.

src/state_variables.jl
bgroenks96: I guess it's fine for now.

docs/src/running/initialization.md
bgroenks96: I wouldn't do it in this PR. It's not directly necessary.

bgroenks96 (approved):

PR #131, opened 2026-06-11
Add initial implementation of heterogeneous soil properties

### LINKED ISSUE #130: Add heterogeneous soil properties with input data provided by SoilGrids 2.0
I just realized that I never created an issue to track this.

As a first step towards incorporating spatial heterogeneity in global simulations, we need to allow for spatially varying soil properties. This was originally planned for v0.1 but pushed back due to time constraints.

This work consists of basically two parts:
- [ ] Refactoring soil stratigraphy types to allow for both spatially variable soil properties; in the vertical direction, this should be accomplished using homogeneous chunks called "soil horizons".
- [x] Incorporating soil input data via `NumericalEarth` (see NumericalEarth/NumericalEarth.jl#332)

bgroenks96:
This PR replaces `HomogeneousStratigraphy` with a more general type `SoilStratigraphy` and convenience constructor `HomogeneousSoilStratigraphy`.

`SoilStragigraphy` now holds a `NamedTuple` of `AbstractSoilHorizon`s vertically ordered from top to bottom. There are currently two implementations of `AbstractSoilHorizon`:
- `ConstantSoilHorizon` which defines a vertical layer with laterally and temporally constant material properties that are treated as parameters
- `PrescribedSoilHorizon` which defines the mineral texture and horizon thickness as `input` variables

TODO:
- [x] Add example using `NumericalEarth` to load SoilGrids 2.0 data
- [x] Update docs

Resolves #130


src/processes/soil/stratigraphy/soil_horizon.jl
```diff
@@ -0,0 +1,63 @@
...
+        ::Type{NF};
+        porosity = ConstantSoilPorosity(NF),
+    ) where {NF}
+    return PrescribedSoilHorizon(porosity)
+end
+
+variables(horizon::PrescribedSoilHorizon{NF}) where {NF} = (
+    input(:sand, XY(), default = one(NF), bounds = UnitInterval, desc = "Mass fraction of sand in soil matrix"),
+    input(:silt, XY(), bounds = UnitInterval, desc = "Mass fraction of silt in soil matrix"),
+    input(:clay, XY(), bounds = UnitInterval, desc = "Mass fraction of clay in soil matrix"),
```
maximilian-gelbrecht: Stupid and naive question. Are those three the main types? I somehow expected more different types without knowing anything about this. So is this just a simplification or is this also what complex models use?

src/processes/soil/stratigraphy/soil_stratigraphy.jl
```diff
@@ -0,0 +1,175 @@
+"""
+    $TYPEDEF
+
+Represents a soil stratigraphy as a stack of named [soil horizons](https://en.wikipedia.org/wiki/Soil_horizon).
+Each soil horizon is assumed to have internally homogeneous soil properties. The number of horizons and their
+respective names are defined by the user.
```
maximilian-gelbrecht, line 6: Names are like below e.g. bedrock, surface, etc, right?

src/processes/soil/stratigraphy/soil_stratigraphy.jl
```diff
@@ -0,0 +1,175 @@
...
+
+Base.length(strat::SoilStratigraphy) = length(strat.horizons)
+Base.iterate(strat::SoilStratigraphy) = Base.iterate(strat.horizons)
+Base.iterate(strat::SoilStratigraphy, iter) = Base.iterate(strat.horizons, iter)
+
+# Variables
+
+function variables(strat::SoilStratigraphy)
+    horizon_vars = map(variables, strat.horizons)
+    return namespaces(horizon_vars)
```
maximilian-gelbrecht: Ah here are the namespaces, I was wondering about this.

src/processes/soil/stratigraphy/soil_stratigraphy.jl
```diff
@@ -0,0 +1,175 @@
...
+# Kernel functions
+
+@inline function soil_horizon(i, j, k, grid, fields, strat::SoilStratigraphy{NF}) where {NF}
+    fgrid = get_field_grid(grid)
+    # get midpoint of current node at k
+    zₖ = znode(i, j, k, fgrid, Center(), Center(), Center())
+    # initially set z to uppermost layer boundary
+    z = znode(i, j, fgrid.Nz + 1, fgrid, Center(), Center(), Face())
+    iₖ = 1
+    for (l, name) in enumerate(keys(strat.horizons))
```
maximilian-gelbrecht: To wrap the discussion we had about this on Tuesday: I think it's totally fine to do it like this. With just a handful of horizons that's also really not that bad on CPU either.

maximilian-gelbrecht (approved):
Really nice! Great implementation. This was also easy for me to follow.

src/processes/soil/stratigraphy/soil_horizon.jl
bgroenks96: Yes, these are the three main components of soil texture, i.e. mineral solid constituents. People sometimes also include gravel which is loose, coarse fragments that are greater than 2 mm in diameter. You might have been thinking of soil *types* which are much more broad. Texture is one characteristic that goes into classifying soils.

src/processes/soil/stratigraphy/soil_stratigraphy.jl
bgroenks96: Yes

PR #133, opened 2026-06-12
Input with Namespaces

maximilian-gelbrecht:
# Summary of what's done here

## New functions

- `inputpath(name)` *(exported)* — normalizes `:x` / `:ns => :x` / `(:ns, :x)` to a path
  tuple `(namespace..., varname)`; also `inputpath(::InputSource)`.
- `matches_scope(source, scope)` — whether a source targets the namespace at `scope`.
- `namespaced_variables(path, var)` — wraps an input declaration in nested `Namespace`s.
- `merge_namespaces(namespaces)` — merges same-named `Namespace`s (used by `Variables`).
- `get_fields(state, ns::Namespace)` — resolves all fields declared by a namespace.
- `soil_horizon_index(i, j, k, ...)` + `horizon_fields(fields, strat, l)` — stratigraphy
  kernel helpers.
- `normalize_texture!(sand, silt, clay; defaults)` *(exported)* — normalize texture
  fractions to sum to 1, fill invalid (NaN) cells with defaults.
- `get_field_grid(::Oceananigans.AbstractGrid) = grid` identity method.

## Changed

- `InputSource` constructors accept namespaced `name`s (Symbol / Pair chain / tuple); the
  `name` type parameter now stores the normalized path tuple; `varname` returns its last
  entry.
- `initialize!`/`update_inputs!` (sources + `StateVariables`) take a trailing
  `scope::InputPath = ()`; the namespace recursion threads the path so sources only write
  into their target namespace. Same in `RasterInputSource` (Rasters extension).
- `variables(source)` returns `Namespace`-wrapped input declarations.
- `Variables(...)` merges same-named namespaces instead of silently dropping duplicates;
  `deduplicate_vars` accepts `Namespace`s.
- `get_fields(state, vars)` resolves `Namespace`s into a nested `namespaces` key, so
  kernels see per-horizon fields (mirrors the `StateVariables` layout).
- Stratigraphy kernels (`soil_texture`/`organic_fraction`/`porosity`) read texture via the
  horizon's namespace fields — `PrescribedSoilHorizon` is now actually usable (it has no
  `.texture` field).
- `AbstractStratigraphy <: AbstractProcess` so its namespaced variables are collected by
  `variables(model)` (they were silently dropped before).
- Soil hydrology `get_fields` call sites now include `strat` (saturation/pressure
  closures, `compute_hydraulics!`, RRE auxiliary/tendencies); the energy path already did.
- Horizon structs use `@parameterized @kwdef` (bare `@parameterized` no longer compiles
  with the current SpeedyWeatherInternals).
- `initialize!(fields, ::FieldInputSource, clock::Clock = Clock(time = 0.0), ...)` — no
  `nothing` default (fixed a method ambiguity with `initialize!(state, grid, ::Nothing, ...)`).
- Oceananigans compat extended to 0.111 (manifest resolves 0.110.1) so NumericalEarth can
  coexist.

## Tests and examples

- New `test/inputs/namespaced_inputs.jl`: `inputpath` normalization, scoped
  initialize/update isolation (root vs. sibling namespaces), namespaced
  `FieldTimeSeries` updates, namespace merging, full `SoilModel` + `OARSoilStratigraphy`
  integration.
- Wired previously-unincluded `soil_stratigrapy_tests.jl` and `soil_composition_tests.jl`
  into the suite; added a `normalize_texture!` test.
- New example `examples/simulations/soil_heat_global_soilgrids.jl`: SoilGrids 2.0 texture
  via NumericalEarth.jl → per-horizon namespaced inputs (`:organic => :sand`, ...) →
  `OARSoilStratigraphy`. The full SoilGrids2 → namespaced-input → model pipeline was
  validated against the real ~2.6 GB dataset: correct geographic orientation, distinct
  per-horizon textures, multiple timesteps run cleanly.
- `NumericalEarth` added to `examples/Project.toml`, pinned to `main` via `[sources.NumericalEarth]`
  (the `SoilGrids` submodule is not in the registered release; it requires Oceananigans
  ≥ 0.110.1, matching the compat bump).

## Notes for whoever runs the SoilGrids example

- Needs `NumericalEarth#main` (registered v0.5.5 has no `SoilGrids`).
- Data download is ~2.6 GB on first use (cached in a Julia scratchspace).
- `Field(Metadatum(...))` returns a `(3956, 1979, 6)` array whose longitude axis is rolled
  180° relative to its `λnodes` (column 1 = prime meridian); latitude is south→north. The
  example's `soilgrids_sampler` handles this with direct nearest-neighbor sampling — the
  native grid is not a valid RingGrids `FullClenshawGrid`, so do not wrap it as one.


src/processes/soil/stratigraphy/soil_texture.jl
```diff
@@ -33,6 +33,29 @@ function Oceananigans.Fields.set!(fields, texture::SoilTexture)
...
 end
 
+"""
+    $SIGNATURES
+
+Normalize the given `sand`, `silt`, and `clay` fraction arrays (or `Field`s) in place such that
+the fractions sum to unity in each element, as required by the `SoilTexture` constructor. Elements
+without valid data (`NaN` or non-positive total) are filled with the `defaults`. This is useful
+for preprocessing soil texture data from external sources (e.g. SoilGrids) where the fractions
+are predicted independently and thus do not sum exactly to unity.
```
bgroenks96, line 43: That's actually not good... probably we need to use `ConservativeRegridding` on the soil texture data.

src/input_output/input_sources.jl
```diff
@@ -20,27 +20,67 @@ Returns a tuple of `Symbol`s corresponding to variable names supported by this `
 """
 variables(::InputSource) = ()
 
+"""
+Type alias for namespaced input variable paths of the form `(namespace_1, ..., namespace_N, varname)`.
+"""
+const InputPath = Tuple{Vararg{Symbol}}
```
bgroenks96, line 26: I like the general idea here, but I think it probably doesn't need to be limited to just inputs.

bgroenks96 (commented):
I like the general approach here, but there are a lot of extraneous changes that overlap with the updated state of my branch. So I think what I will do is just close this PR but adapt some of the code and the example into the original PR.

maximilian-gelbrecht:
> I like the general approach here, but there are a lot of extraneous changes that overlap with the updated state of my branch. So I think what I will do is just close this PR but adapt some of the code and the example into the original PR.

Yes, fine with me!

bgroenks96:
I actually don't think the `InputPath` is necessary. The previous implementation already recurses over namespaces, so technically it would have worked (my bad). But I think the basic functionality of defining paths through namesapces could be useful, so I'm going to keep it.

Issue #126, opened 2026-06-08
Terrarium fails on undefined Thermodynamics.parameters


pierluigividale:
I have built the latest versions of SpeedyWeather, Oceanigans and Terrarium.
I am trying the example under Land, where Terrarium is defined and then initialised as coupled to Speedy.

The model fails instantly because it cannot find Rv_over_Rd. See the error pasted below.

Looking at the code in Terrarium, there are definitions for Thermodynamics.constants and Thermodynamics.parameters, but the way that UniversalConstants is defined includes Termodynamic.constants, albeit not Thermodynamics.parameters. Is this why the model fails?

```UniversalConstants(::Type{NF}; kwargs...) where {NF} = UniversalConstants{NF}(; kwargs...)

"""
    $TYPEDEF

Top-level container for all physical constants used in Terrarium. Groups three
sub-structs by category:

- [`ThermodynamicConstants`](@ref) — thermodynamic and atmospheric constants
- [`MaterialConstants`](@ref) — material properties of water, ice, and carbon
- [`UniversalConstants`](@ref) — universal constants (gravity, Stefan-Boltzmann, von Kármán)
```

Here is what I see after I load Terrarium, indicating that Thermodynamics.parameters is indeed missing:
```
ulia> using Terrarium
[ Info: Oceananigans will use 10 threads

julia> show(PhysicalConstants())
PhysicalConstants{Float64}(ThermodynamicConstants{Float64}(1004.5, 2070.0, 4186.0, 1846.0, 334000.0, 2.257e6, 2.834e6, 273.16, 273.16, 273.16, 611.657, 287.058, 461.5), MaterialConstants{Float64}(1000.0, 916.7, 12.0), UniversalConstants{Float64}(9.80665, 5.6704e-8, 0.4))
```

Here is the runtime error:
```
ERROR: LoadError: UndefVarError: `Rv_over_Rd` not defined in `Thermodynamics.Parameters`
Suggestion: check for spelling errors or missing imports.
Stacktrace:
  [1] getproperty(x::Module, f::Symbol)
    @ Base ./Base_compiler.jl:47
  [2] top-level scope
    @ ~/.julia/packages/Terrarium/dGpyX/src/processes/physical_constants.jl:160
  [3] include(mapexpr::Function, mod::Module, _path::String)
    @ Base ./Base.jl:307
  [4] top-level scope
    @ ~/.julia/packages/Terrarium/dGpyX/src/processes/processes.jl:4
[...]

PR #135, opened 2026-06-15
Investigate and fix possible bugs in surface energy balance calculations

### LINKED ISSUE #132: Surface energy balance: iteration and coupling to surface hydrology
Based on the [land_column example](https://github.com/NumericalEarth/Terrarium.jl/blob/34fa7f26eb3878095847f13c1e648f2094464601/examples/simulations/land_column.jl) I tried to assess the surface energy balance code for the coupled land model. 

How this currently goes:

- Skin temperature (T_s) is set to 0°C by default
- The surface hydrology model is the first to calculate its auxiliary variables based on this T_s. It calculates tranpsiration, ground evaporation and canopy evaporation: https://github.com/NumericalEarth/Terrarium.jl/blob/34fa7f26eb3878095847f13c1e648f2094464601/src/models/coupled/land_model.jl#L84 
- Then the surface energy balance gets its auxiliary variables calculated. https://github.com/NumericalEarth/Terrarium.jl/blob/34fa7f26eb3878095847f13c1e648f2094464601/src/models/coupled/land_model.jl#L85 The problem is that the latent heat flux here is seen as the sum of the surface hydrology outputs (the three components mentioned above), as they are always summed in https://github.com/NumericalEarth/Terrarium.jl/blob/34fa7f26eb3878095847f13c1e648f2094464601/src/processes/surface_energy/surface_energy_balance.jl#L139 As a consequence, when doing the 1 iteration for T_s, LE is considered fixed. Only G,H and R_n can change based on the new values for T_s.

Additionally, I am not sure if this 1 iteration of the surface energy balance will actually converge, especially since you start from this 0°C. Then at the next time step, you start from T_s from the previous, but at no point are we are sure of convergence. 

I think it makes more sense to do a full nonlinear solve, as e.g. outlined in Bonan 2019: 

Image

This is not trivial however. As bonan mentions, the question is also if you then couple then soil temperature itself in this iteration or not 

See also #126 and SpeedyWeather/SpeedyWeather.jl#1123

bgroenks96:
This PR aims to investigate address several potential issues in the existing surface energy balance scheme highlighted in #132.

## Tasks
- [x] Review formulations of bulk fluxes
- [x] Review integration of Thermodynamics.jl for potential issues (@olivierbonte)
- [x] Add tests for thermodynamics functions (@olivierbonte)
- [x] Use more realistic input values in skin temperature tests
- [x] Investigate validity of single fixed-point iteration
- [x] Add option for doing an iterative solve (not just one step) in SEB
- [x] Fix coupling with surface hydrology

Resolves #132

See also https://github.com/NumericalEarth/Terrarium.jl/issues/126 and https://github.com/SpeedyWeather/SpeedyWeather.jl/issues/1123


olivierbonte:
@bgroenks96 It seems like the clima function for calculation saturated vapor specific humidity is given very low values which can't be correct. I added a test for the thermodynamics which now checks this. I'll see if this is my mistake or upstream.

olivierbonte:
Okay, nevermind, I made a mistake in the tests, I'll check some more but I don't think the thermodynamics is the issue for now

olivierbonte:
With the updated example of sunny and dry, I now get 
```
skin temperature at iteration 0 (default Terrarium): 19.318596136871513 residual: 0.0  energy balance: 2.842170943040401e-14 net radiation: -308.5037966563586 Hₗ: 115.54294547112373 Hₛ: 70.09587524473937 G: -122.86497594049548
skin temperature at iteration 1: 15.917327943837813  residual: 3.401268193033699  energy balance: 0.0 net radiation: -334.19245446460775 Hₗ: 66.68716053046536 Hₛ: 34.119058427117345 G: -233.38623550702505
```
to then end up at 
```
skin temperature at iteration 20: 15.618353649819198  residual: 0.0  energy balance: 2.842170943040401e-14 net radiation: -328.488959674753 Hₗ: 76.7585704018065 Hₛ: 42.262097287410626 G: -209.46829198553593
```

The ground heat flux is very confusing to me. So when skin temperature is 19.3°C, the ground heat flux is lower in magnitude then when it is at 15.6 °C. This should not be the case, as the ground temperature is 13°C and so the gradient is larger at the first iteration.

olivierbonte:
Also I think that the fixed thermal conductivity at 2 W/(mK) for the soil is a bit on the high side as a default 

image

When setting it to 0.5, the iterative calculation crashed.

bgroenks96:
Hey @oliverbonte, thanks for already looking into this. I added a relaxation step to the fixed-point update which should resolve the crash you saw for lower conductivity. We should really update this to use the actual soil conductivity (when provided), though... I completely forgot about this.

> The ground heat flux is very confusing to me. So when skin temperature is 19.3°C, the ground heat flux is lower in magnitude then when it is at 15.6 °C. This should not be the case, as the ground temperature is 13°C and so the gradient is larger at the first iteration.

The ground heat flux here is calculated as the residual of the other fluxes, not directly from the temperature gradient. This gradient is used to calculate the updated skin temperature.

> Also I think that the fixed thermal conductivity at 2 W/(mK) for the soil is a bit on the high side as a default

I changed the default value to 1 W/(m K) for now.

olivierbonte:
Thanks for updating this already!

> The ground heat flux here is calculated as the residual of the other fluxes, not directly from the temperature gradient. This gradient is used to calculate the updated skin temperature.
> 

Okay, but once converged, the gradient formulation should equal the residual right? I added a test for this and after convergence this does match

bgroenks96:
Yes they should match after convergence.

olivierbonte:
Setting the aerodynamic conductance a bit higher in https://github.com/NumericalEarth/Terrarium.jl/pull/135/commits/b44a007b9a41931e5fdf439dbe6d5de1d454eb2b gives more realistic values of fluxes. However, no the default 1 iteration of Terrarium gives 
```
skin temperature at iteration 0 (default Terrarium): 21.22000420520847 net radiation: -297.934751671891 Hₗ: 306.6842979148376 Hₛ: 140.67624001554796 G: 149.42578625849458
```
And convergence after 20 iterations
```
skin temperature at iteration 20: 16.314048163535286  residual: 3.581490659598785e-11 net radiation: -324.78959642793376 Hₗ: 179.34319447459058 Hₛ: 79.16543868188604 G: -66.28096327145715
```

So in this case, the current approach of 1 iteration does not work, and a fixed point iterations seems necessary to arrive at a reasonable value of skin temperature

test/thermodynamics_test.jl
```diff
@@ -19,8 +21,8 @@ using Test
     e_sat_terrarium = Terrarium.saturation_vapor_pressure(thermodyn_constants, air_temperature)
     q_sat_terrarium = Terrarium.saturation_specific_humidity_vapor(thermodyn_constants, air_temperature, air_density)
     # Compare
-    @test e_sat_clima ≈ e_sat_terrarium
-    @test q_sat_clima ≈ q_sat_terrarium
+    @test isapprox(e_sat_clima, e_sat_terrarium; rtol = 0.01)
```
bgroenks96: @olivierbonte Do you know why this is necessary?

bgroenks96:
> So in this case, the current approach of 1 iteration does not work, and a fixed point iterations seems necessary to arive at a reasonable value of skin temperature

Ok.. then so be it. I'll add an option to iterate it.

test/thermodynamics_test.jl
olivierbonte: This is because our constant are not 100% the same as those in  [ClimaParams.jl](https://github.com/CliMA/ClimaParams.jl/blob/7c11c6525d1bc2d4166004ff45a685916de04953/src/parameters.toml). If you want, I can adapt them to be exactly the same, and then we can  test using  ≈

bgroenks96: Hmm, which ones are different?

olivierbonte: - Isobaric specific heat capacity of water vapour: current = 1846.0, Climapars = 1859
- Isobaric specific heat capacity of liquid water: current = 4186, Climaparms = 4181
- Freezing temperature water: current = 273.16, Climaparams = 273.15
- Latent heat of sublimation: current = 2.834e6, ClimaParams = 2.8344e6
- gas constant dry air: current = 287.058 ClimaParams = 287.0 

I see no harm in making these 100% consistent, what do you think?

bgroenks96: Yeah I would make them consistent. Not sure why the freezing temperature is 273.16... seems like a typo.

olivierbonte: Alright, this should now be handled in https://github.com/NumericalEarth/Terrarium.jl/pull/135/commits/9c51c5fefe7d931d908b29aa277f910085211e91

bgroenks96:
This is not fully tested yet, and I'm guessing there are still some broken unit tests, but I think it's ready for review @olivierbonte.

In the end, I had to implement a very naive Newton-Raphson solver to make it work reliably. The fixed-point iteration just wasn't stable enough. If you want, you can refactor this in the future to use RootSolvers.jl.

olivierbonte:
I'll try to have a look at this soon @bgroenks96! 

I'll think a bit more about whether or not to refactor this to RootSolvers.jl at some point, not sure yet what the right balance of implementing ourselves vs reuse of other packages is here.

bgroenks96:
I looked at RootSolvers.jl again and it seems simple enough to use. I think I'll try just replacing the solvers in this PR with that and see what happens. I vibe coded the Newton-Raphson solver here, so I'm not attached to it at all :sweat_smile:

bgroenks96:
@olivierbonte [fe29b97](https://github.com/NumericalEarth/Terrarium.jl/pull/135/commits/fe29b9778bc2ad3bbcf71e5022cae799e9cf19af) adds a wrapper for RootSolvers.jl. It seems to be a bit slower at the moment though than my vibe-coded one, interestingly.

src/processes/surface_energy/skin_temperature.jl
```diff
@@ -37,21 +37,32 @@ variables(::PrescribedSkinTemperature) = (
...
+    solver::Solver
 end
 
-ImplicitSkinTemperature(::Type{NF}; kwargs...) where {NF} = ImplicitSkinTemperature{NF}(; kwargs...)
+ImplicitSkinTemperature(::Type{NF}; κₛ::NF = NF(1.0), solver = default_skin_temperature_solver(NF)) where {NF} = ImplicitSkinTemperature{NF, typeof(solver)}(κₛ, solver)
+
+function default_skin_temperature_solver(::Type{NF}) where {NF}
+    # Default to Newton-Raphson without relaxation (factor = 1) but clamping values to realistic limits
+    relax = RelaxationFactor(factor = NF(1.0), upper_limit = NF(80), lower_limit = NF(-80))
+    return NewtonRaphsonSolver(NF, max_iterations = 5, relax = relax)
```
olivierbonte: Isn't 5 a bit on the low end? Wouldn't it be better here to a set a absolute convergence criterion? Bonan in his book suggestes 0.001K

src/processes/surface_energy/skin_temperature.jl
```diff
@@ -70,11 +81,14 @@ end
...
-Compute the ground heat flux as the residual of the net radiation `R_net` and the
-sensible `H_s` and latent `H_l` heat flux.
+Compute the ground heat flux `G` that closes the surface energy balance. With all fluxes
+positive upward (aligned with `+z`), the energy arriving at the skin from below must balance
+the radiative and turbulent losses above, so `G = R_net + H_s + H_l`.
 """
 @inline function compute_ground_heat_flux(::AbstractSkinTemperature, R_net, H_s, H_l)
-    # Compute ground heat flux as the residual of R_net and turbulent fluxes
+    # Surface energy balance with all fluxes positive upward (aligned with +z):
+    # energy reaching the skin from below (G) must balance the radiative and turbulent
```
olivierbonte: Already in docstrings above, so maybe better to avoid the duplication?

src/processes/surface_energy/skin_temperature.jl
```diff
@@ -170,12 +183,35 @@ Compute the ground heat flux from the surface net radiation and sensible/latent
     return G
 end
 
+@propagate_inbounds function update_skin_temperature!(
```
olivierbonte: This no has no docstrings anymore -> replace update_skin_temperature! by compute_skin_temperature! in the docs

docs/src/processes/surface_energy/surface_energy_balance.md
```diff
@@ -46,6 +46,12 @@ compute_auxiliary!(state, grid, seb::SurfaceEnergyBalance, constants::PhysicalCo
...
 
+The `SurfaceEnergyBalance` process provides two primary method interfaces:
+
+```@docs; canonical = false
+compute_surface_energy_fluxes!(state, grid, seb::SurfaceEnergyBalance, constants::PhysicalConstants, atmos::AbstractAtmosphere, hydrology::Optional{AbstractSurfaceHydrology}, args...)
+```
+
 ```@docs; canonical = false
-compute_surface_energy_fluxes!
+solve_surface_energy_balance!(state, grid, seb::SurfaceEnergyBalance, constants::PhysicalConstants, atmos::AbstractAtmosphere, hydrology::Optional{AbstractSurfaceHydrology}, args...)
```
olivierbonte: change seb::SurfaceEnergyBalance to seb::SurfaceEnergyBalance{NF, <:ImplicitSkinTemperature}, I think this might prevent the docs from failing

src/processes/surface_energy/skin_temperature.jl
```diff
@@ -170,12 +183,35 @@ Compute the ground heat flux from the surface net radiation and sensible/latent
...
+    return nothing
+end
+
+@propagate_inbounds function solve_skin_temperature!(
+        out, i, j, grid, fields,
+        skinT::ImplicitSkinTemperature,
+        seb::AbstractSurfaceEnergyBalance,
+        seb_args...
+    )
+    objective = ObjectiveFunction(update_skin_temperature!, :skin_temperature)
```
olivierbonte: The objective function now is a function of the form typical for fixed point iteration: x_{n+1} = f(x_n). 

In general, I think is easier however to provide objective functions in terms of residuals: f(x) = res = 0?

src/processes/surface_hydrology/evapotranspiration/bare_ground_evaporation.jl
```diff
@@ -21,9 +21,31 @@ BareGroundEvaporation(
 
 @propagate_inbounds surface_humidity_flux(i, j, grid, fields, evaporation::BareGroundEvaporation, args...) = fields.evaporation_ground[i, j]
 
+@inline ground_evaporation_conductance(ET::AbstractEvapotranspiration, β, rₐ) = β / rₐ
```
olivierbonte, line 25: I get the rationale, but a bit confusion that it is not the inverse of the ground_evaporation_conductance because you are also taking into account the aerodynamic term. maybe this can be clarified by e.g. adding "total" of "effective" as extra suffix?

src/processes/surface_hydrology/evapotranspiration/bare_ground_evaporation.jl
```diff
@@ -21,9 +21,31 @@ BareGroundEvaporation(
...
+    Ts = fields.skin_temperature[i, j]
+    g = fields.ground_evaporation_conductance[i, j]
+    Δq = compute_specific_humidity_difference(i, j, grid, fields, atmos, constants, Ts)
+    return g * Δq
+end
+
 # Top-level interface methods
 
 variables(::BareGroundEvaporation) = (
+    # Skin-driven vapor conductance β/rₐ (independent of skin temperature; held fixed during the SEB solve)
```
olivierbonte, line 48: In the current formulation, both beta (Pielke) and r_a (fixed drag coefficient), are independent of T_s.

Once your r_a becomes dependent on T_s because of stability effects, you don't want this to be fixed during the solve right? see e.g. Bonan

Image

src/solvers/fixed_point.jl
```diff
@@ -0,0 +1,44 @@
+struct FixedPointSolver{NF, R}
```
bgroenks96, line 12: We should decide @olivierbonte whether to keep this and the `NewtonRaphson` solver in the codebase or not. The `RootSolvers` wrapper makes at least the latter kind of redundant.

olivierbonte: Agree, I won't have time to finish the code review today, but I'll get back at it tomorrow morning. My gut feeling would say drop the custom NewtonRaphson.

src/processes/surface_energy/surface_energy_balance.jl
olivierbonte, line 47: Okay for me to for now keep the nonlinearsolve inside of compute_auxiliary!, but as mentioned in https://github.com/NumericalEarth/Terrarium.jl/issues/132#issuecomment-4707089792 we should consider making an extra method for this

src/processes/surface_energy/surface_energy_balance.jl
```diff
@@ -131,14 +122,44 @@ Fused kernel function that computes the radiative and turbulent fluxes, as well
...
+        atmos::AbstractAtmosphere,
+        evtr::Optional{AbstractEvapotranspiration} = nothing,
+        args...
+    )
+    i, j = @index(Global, NTuple)
+
+    # Compute fluxes based on current skin temperature
+    solve_skin_temperature!(out, i, j, grid, fields, seb.skin_temperature, seb, constants, atmos, args...)
+    if !isnothing(evtr)
+        # Recompute evaporation component fluxes from final skin temperature
```
olivierbonte: Minor comment, but would be good to land on a choice between "evaporation" or "evapotranspiration" for the combination of transpiration + canopy evaporation + bare soil evaporation. I personally prefer the term "evaporation" (see https://doi.org/10.1029/2020WR028055 for argumentation), but interested to hear what you think.

src/processes/surface_hydrology/evapotranspiration/bare_ground_evaporation.jl
```diff
@@ -46,18 +68,51 @@ end
 
 # Kernel functions
 
-@propagate_inbounds function compute_evapotranspiration!(
+@propagate_inbounds function compute_evapotranspiration_conductance!(
         out, i, j, grid, fields,
         evaporation::BareGroundEvaporation,
```
olivierbonte, line 80: same comment as above: evaporation vs evapotranspirtaion choice

src/processes/surface_hydrology/evapotranspiration/bare_ground_evaporation.jl
olivierbonte: Also maybe not ideal to have beta/r_a as the effective ground_evaporation_conductance for all AbstractEvapotranspiration types. a r_soil parametrisation is equally valid 

Image

src/processes/surface_hydrology/evapotranspiration/bare_ground_evaporation.jl
```diff
@@ -21,9 +21,31 @@ BareGroundEvaporation(
...
+evaporation conductance in `fields`.
+"""
+@propagate_inbounds function compute_surface_humidity_flux(
+        i, j, grid, fields,
+        ::BareGroundEvaporation,
+        atmos::AbstractAtmosphere,
+        constants::PhysicalConstants,
+    )
+    Ts = fields.skin_temperature[i, j]
+    g = fields.ground_evaporation_conductance[i, j]
```
olivierbonte, line 40: related to comment above: g_eff or something similar maybe more clear

src/processes/surface_hydrology/evapotranspiration/canopy_evapotranspiration.jl
```diff
@@ -37,56 +37,34 @@ function PALADYNCanopyEvapotranspiration(
...
-
-Compute evaporation of water intercepted by the canopy from humidity gradient `Δq`, canopy saturation fraction
-`f_can`, and aerodynamic resistance `rₐ`.
-"""
-@inline function compute_evaporation_canopy(::PALADYNCanopyEvapotranspiration, Δq, f_can, rₐ)
-    # Calculate canopy evaporation flux in m/s (positive upwards)
-    E_can = f_can * Δq / rₐ
-    return E_can
-end
+@inline canopy_evaporation_conductance(::PALADYNCanopyEvapotranspiration, f_can, rₐ) = f_can / rₐ
```
olivierbonte, line 58: Also here I would propose a pre of suffix like 'effective' to denote that is it not merely 1/r_a

src/processes/surface_hydrology/evapotranspiration/canopy_evapotranspiration.jl
```diff
@@ -101,6 +79,26 @@ variables(::PALADYNCanopyEvapotranspiration{NF}) where {NF} = (
...
+@propagate_inbounds function compute_surface_humidity_flux(
+        i, j, grid, fields,
+        ::PALADYNCanopyEvapotranspiration,
+        atmos::AbstractAtmosphere,
+        constants::PhysicalConstants,
+    )
+    Ts = fields.skin_temperature[i, j]
+    g_can = fields.canopy_evaporation_conductance[i, j]
+    g_trp = fields.transpiration_conductance[i, j]
+    E_gnd = fields.evaporation_ground[i, j]
```
olivierbonte: Why isn't ground evaporation calculated in this function? 
To me it makes more sense to calculate all 3 the components of evporation here

src/processes/surface_hydrology/evapotranspiration/canopy_evapotranspiration.jl
```diff
@@ -141,18 +139,54 @@ for the given scheme `evapotranspiration` and process dependencies.
     Tg = fields.ground_temperature[i, j] # ground temperature (top snow/soil layer)
     gw_can = fields.canopy_water_conductance[i, j] # stomatal conductance (assumed to be defined by vegetation)
 
-    # Compute VPD and resistance terms
+    # Compute VPD and resistance termsrefreshes
```
olivierbonte: typo?

src/processes/surface_hydrology/evapotranspiration/canopy_evapotranspiration.jl
```diff
@@ -141,18 +139,54 @@ for the given scheme `evapotranspiration` and process dependencies.
...
     Δqs = compute_specific_humidity_difference(i, j, grid, fields, atmos, constants, Ts) # humidity difference between canopy and atmosphere
     Δqg = compute_specific_humidity_difference(i, j, grid, fields, atmos, constants, Tg) # humidity difference between ground and canopy
```
olivierbonte: Why calculate these? they are not stored to my understanding

src/processes/surface_hydrology/evapotranspiration/canopy_evapotranspiration.jl
```diff
@@ -37,56 +37,34 @@ function PALADYNCanopyEvapotranspiration(
...
-    # Calculate canopy evaporation flux in m/s (positive upwards)
-    E_can = f_can * Δq / rₐ
-    return E_can
-end
+@inline canopy_evaporation_conductance(::PALADYNCanopyEvapotranspiration, f_can, rₐ) = f_can / rₐ
 
 # Top-level interface methods
 
 variables(::PALADYNCanopyEvapotranspiration{NF}) where {NF} = (
+    # Skin-driven vapor conductances (independent of skin temperature; held fixed during the SEB solve)
```
olivierbonte, line 63: Same comment as for bare soil, not sure you want to keep these contant necessarily during SEB solve?

src/processes/surface_hydrology/evapotranspiration/canopy_evapotranspiration.jl
```diff
@@ -126,7 +124,7 @@ end
 Compute `transpiration`, `evaporation_ground`, and `evaporation_canopy` fluxes on `grid`
 for the given scheme `evapotranspiration` and process dependencies.
 """
-@propagate_inbounds function compute_evapotranspiration!(
+@propagate_inbounds function compute_evapotranspiration_conductance!(
```
olivierbonte: better to be called conductances? because you are storing 3 and not 1?

src/processes/surface_hydrology/evapotranspiration/evapotranspiration_base.jl
```diff
@@ -1,3 +1,13 @@
+"""
+    $TYPEDSIGNATURES
+
+Compute an evapotranspiration flux [m/s] (positive upwards) as the product of a vapor
+conductance `g` and a specific humidity difference `Δq`. All evapotranspiration components
```
olivierbonte: maye add units for clarity: g [m/s] and deltaq [kg/kg]

src/solvers/fixed_point.jl
```diff
@@ -0,0 +1,44 @@
...
+        tolerance::NF = sqrt(eps(NF)),
+        relax::R = RelaxationFactor(factor = NF(0.5)),
+        max_iterations::Int = 100
+    ) where {NF, R}
+    return FixedPointSolver{NF, R}(tolerance, relax, max_iterations)
+end
+
+@propagate_inbounds function solve!(
+        out, indices::NTuple{N, Integer}, grid, fields,
+        step_func!::ObjectiveFunction{target},
```
olivierbonte: maybe needs some clarifications of what is expected here: a map in the form of $u_{n+1} = G(u_n)$

src/solvers/fixed_point.jl
olivierbonte: FixedPointSolvers seems fine for me to keep. In theory, you could also try to extend RootSolvers.jl (https://clima.github.io/RootSolvers.jl/dev/DeveloperDocs/#Extending-RootSolvers.jl) but I think this a bit of a stretch given that fixed point iteration expects a different kind of function than root solving

src/solvers/root_solvers.jl
```diff
@@ -0,0 +1,115 @@
+import RootSolvers
+
+"""
+    $TYPEDEF
+
+Wrapper for RootSolvers.jl root-finding methods.
+"""
+struct RootSolver{NF, M, S}
+    "Numerical tolerance of the root finding iteration"
```
olivierbonte, line 7: maybe this is a Julia misunderstnanding on my part, but given that you can set the method (one of the options in https://clima.github.io/RootSolvers.jl/dev/GettingStarted/#Available-Root-Finding-Methods?) the constructor, why is it not a field here?

src/solvers/root_solvers.jl
```diff
@@ -0,0 +1,115 @@
...
+    tolerance::NF
```
olivierbonte: The tolerance interface of RootSolvers.jl can maybe be reused? https://clima.github.io/RootSolvers.jl/dev/API/#Tolerance-Types

src/solvers/root_solvers.jl
```diff
@@ -0,0 +1,115 @@
...
+        tolerance::NF = sqrt(eps(NF)),
+        max_iterations::Int = 100,
+        solution_type::S = RootSolvers.VerboseSolution()
+    ) where {NF, S}
+    return RootSolver(NF, method, tolerance, max_iterations, solution_type)
+end
+
+@propagate_inbounds function solve!(
+        out, indices::NTuple{N, Integer}, grid, fields,
+        step_func!::ObjectiveFunction{target},
```
olivierbonte, line 33: I  think it makes more sense to ask for a residual function here as input: $F(u) = 0$

src/solvers/root_solvers.jl
```diff
@@ -0,0 +1,115 @@
...
+        solver::RootSolver{NF, M, S},
+        func_args...;
+        func_kwargs...
+    ) where {NF, N, M, S, target}
+    target_field = getproperty(out, target)
+
+    # Define the residual function F(x) = g(x) - x
+    residual! = build_residual(out, indices, grid, fields, step_func!, target_field, solver, func_args...; func_kwargs...)
```
olivierbonte, line 41: See above, I think it makes more sense to do this outside of this solve; you define as a user the residual function you are interested in solving

src/solvers/root_solvers.jl
```diff
@@ -0,0 +1,115 @@
...
+
+    # Solve for the root using RootSolvers.jl
+    x₀ = target_field[indices...]
+    method = M(x₀)
```
olivierbonte, line 45: I don't think this M is particullary clear, better for it to be a field of the struct (see above)?

src/solvers/root_solvers.jl
```diff
@@ -0,0 +1,115 @@
...
+        return new{NF, M, typeof(solution_type)}(tolerance, max_iterations, solution_type)
+    end
+end
+
+function RootSolver(
+        ::Type{NF};
+        method = RootSolvers.NewtonsMethod{NF},
+        tolerance::NF = sqrt(eps(NF)),
+        max_iterations::Int = 100,
+        solution_type::S = RootSolvers.VerboseSolution()
```
olivierbonte: Set Compactsolution as default: only this one is GPU compatible (see https://clima.github.io/RootSolvers.jl/dev/GettingStarted/#GPU-Usage-Tips)

src/solvers/root_solvers.jl
```diff
@@ -0,0 +1,115 @@
...
+    ) where {NF, F}
+
+    function residual!(x)
+        target_field[indices...] = x
+        step_func!(out, indices..., grid, fields, func_args...; func_kwargs...)
+        gx = target_field[indices...]
+        return gx - x
+    end
+
+    function residual_with_derivative!(x)
```
olivierbonte, line 109: not used?

src/solvers/root_solvers.jl
```diff
@@ -0,0 +1,115 @@
...
+    iteration = S <: RootSolvers.CompactSolution ? nothing : result.iter_performed
+
+    # Ensure the target field holds the final estimate
+    target_field[indices...] = x_root
+
+    return x_root, iteration
+end
+
+
+@propagate_inbounds function build_residual(
```
olivierbonte, line 71: As mentioned above, to me it makes more sense to just have a function defined e.g. compute_surface_energy_balance_residual in the main code, which can then be used in the rootsolving.

src/solvers/newton_raphson.jl
olivierbonte, line 1: I would drop this and use RootSolvers.jl

src/solvers/solvers.jl
```diff
@@ -0,0 +1,70 @@
+"""
+    $TYPEDEF
+
+Represents an objective function for nonlinear solvers. The name `target`
+refers to the output `Field` which should updated on each iteration.
+"""
+struct ObjectiveFunction{target, F, DF}
```
olivierbonte, line 15: Define more clearly what an objective function should do. Now it expect something of the form $u = G(u)$.
I think it is a general easier to write an objective function in the form of $F(u) = 0$, as it doesn't require separating out $u$. 

So instead of build_residual, you would have a build_step_function (or build_map) that follows (got this from [the SUNDIALS paper](https://doi.org/10.1145/3539801)):
Nonlinear system: $F(u) = 0$
OR
Fixed-point form: $G(u) =u$
To go from $F(u)$ to $G(u)$, you write
$$F(u) = Lu - N(u) = 0$$
where $L$ is a linear operator and $N$ nonlinear. Then for a fixed point iterations you say:
$$u = L^{-1}N(u)$$
Because $N(u)=Lu - F(u)$ you can also write the above as:
$$u = L^{-1}(Lu - F(u)) = u - L^1F(u)=G(u)$$ 

In the current implementation:
$$F(T_s) = H_l(T_s) + H_s(T_s) + R_n(T_s) - G(T_s) = 0$$
where $G(T_s)$ is the linear part ($L(u)$):
$$L(T_s) = G(T_s) = \kappa \frac{T_g - T_s}{\Delta z}$$

test/surface_hydrology/evapotranspiration_tests.jl
```diff
@@ -0,0 +1,281 @@
+using Terrarium
+using Terrarium:
+    compute_evaporation_flux,
+    transpiration_conductance,
+    canopy_evaporation_conductance,
+    ground_evaporation_resistance_factor
+using Test
+
+# ==============================================================================
```
olivierbonte, line 10: These === are maybe not really that was there before, not sure if you want to adopt this more broadly

Project.toml
```diff
@@ -18,6 +18,7 @@ KernelAbstractions = "63c18a36-062a-441e-b654-da1e3ab1ce7c"
 ModelParameters = "4744a3fa-6c31-4707-899e-a3298e4618ad"
 Oceananigans = "9e8cae18-63c1-5223-a75c-80ca9d6e9a09"
 RingGrids = "d1845624-ad4f-453b-8ff4-a8db365bf3a7"
+RootSolvers = "7181ea78-2dcb-4de3-ab41-2b8ab5a31e74"
```
olivierbonte, line 21: maybe also needs a compat?

olivierbonte (commented):
looks good in general! My main suggestion is to switch to solving residuals instead of solving fixed point maps as a default case

src/processes/surface_energy/skin_temperature.jl
bgroenks96: I think for GPU performance we should avoid a situation where we have lots of GPU threads doing extended nonlinear solves with different numbers of iterations. The skin temperature solution is unlikely to be the biggest source of error in the model, so I don't think it's worth paying such penalties for a more exact solution.

src/solvers/solvers.jl
bgroenks96: Since we are dealing here only with scalar rootfinding problems, it's probably easier to simply reformulate the residual $F(u) = 0$ as $G(u) = u - \alpha F(u)$. Or we can just get rid of the fixed-point solver entirely. It doesn't really serve much of a purpose now that we have `RootSolvers`.

src/processes/surface_energy/surface_energy_balance.jl
bgroenks96: It should be "evaporation" for the flux terms representing direct phase change of free-standing or pore-bound liquid water to vapor, and "evapotranspiration" when also including fluxes due to transpiration. Here it should be evapotranspiration.

I don't find the arguments of Miralles et al. convincing. It seems like a fairly petty and purely semantic discussion to me. Yes, sure, technically transpiration is also a form of evaporation, and yes, it is driven by some of the same hydrometeorological factors. However, transpiration also involves a lot of other physical processes (e.g. plant hydraulics, stomatal conductance, etc), and from a hydrological perspective, clearly constitutes a separate flux contribution since it is conditional on the presence of vegetation in the first place. It also represents a distinct physical pathway by which water can be removed from the soil vs. direct evaporation.

So the term "evapotranspiration" is useful in the sense that it clearly indicates that we are lumping together at least two distinct fluxes that have different physical origins. Ideally, we would also distinguish evaporation of free-standing water from evaporation of pore-bound water, but this is semantically less important.

src/solvers/root_solvers.jl
bgroenks96: Because RootSolvers made the strange API decision of needing to supply the initial value to the method constructor. So we instead store the method type (see type parameter `M`) and then construct it lazily.

src/solvers/root_solvers.jl
bgroenks96: See answer above.

src/solvers/root_solvers.jl
bgroenks96: This seems like a duplicate of the many other comments regarding fixed-point vs. residual formulation.

src/solvers/root_solvers.jl
bgroenks96: It is used, see the return statement.

src/processes/surface_energy/skin_temperature.jl
olivierbonte: Okay, that makes sense from a computatoinal perspective. 
Let's keep it this way then if you prefer that, it is only a default of course, it can still be changed for e.g. site runs

src/processes/surface_energy/surface_energy_balance.jl
olivierbonte: No sure, all valid arguments, in [Ahmad et al. (2025)](https://doi.org/10.1029/2024WR038835) a similar reasoning is set up. I also think it is only a semantic dicscussion. "Evapotranspiration" as "evaporation" (no physiology, canopy + ground) and "transpiration" (including physiology) for sure also works.

src/solvers/root_solvers.jl
olivierbonte: Ahn okay, thanks clarifying!

src/solvers/root_solvers.jl
olivierbonte: oops sorry, missed that

src/solvers/solvers.jl
olivierbonte: Don't you have to tune the $\alpha$ then be sure the fixed point solver converges?
For me also fine to drop the fixed point solver!

src/processes/surface_energy/skin_temperature.jl
bgroenks96: Done in [560a0fb](https://github.com/NumericalEarth/Terrarium.jl/pull/135/commits/560a0fb1d3b1530d0ef7c708f8582dd0505377c5)

src/processes/surface_energy/skin_temperature.jl
bgroenks96: Done in [560a0fb](https://github.com/NumericalEarth/Terrarium.jl/pull/135/commits/560a0fb1d3b1530d0ef7c708f8582dd0505377c5)

docs/src/processes/surface_energy/surface_energy_balance.md
bgroenks96: Fixed in [560a0fb](https://github.com/NumericalEarth/Terrarium.jl/pull/135/commits/560a0fb1d3b1530d0ef7c708f8582dd0505377c5)

src/processes/surface_energy/skin_temperature.jl
bgroenks96: Done in [560a0fb](https://github.com/NumericalEarth/Terrarium.jl/pull/135/commits/560a0fb1d3b1530d0ef7c708f8582dd0505377c5)

src/solvers/fixed_point.jl
bgroenks96: I guess this is addressed in  [560a0fb](https://github.com/NumericalEarth/Terrarium.jl/pull/135/commits/560a0fb1d3b1530d0ef7c708f8582dd0505377c5)

src/solvers/root_solvers.jl
bgroenks96: Done in [560a0fb](https://github.com/NumericalEarth/Terrarium.jl/pull/135/commits/560a0fb1d3b1530d0ef7c708f8582dd0505377c5)

src/solvers/newton_raphson.jl
bgroenks96: Done in [560a0fb](https://github.com/NumericalEarth/Terrarium.jl/pull/135/commits/560a0fb1d3b1530d0ef7c708f8582dd0505377c5)

src/processes/surface_hydrology/evapotranspiration/bare_ground_evaporation.jl
bgroenks96: > I get the rationale, but a bit confusion that it is not the inverse of the ground_evaporation_conductance because you are also taking into account the aerodynamic term.

I am not sure I understand your complaint here. In what formulation of the ground evaporation term is $r_a$ not accounted for?

> Also maybe not ideal to have beta/r_a as the effective ground_evaporation_conductance for all AbstractEvapotranspiration types. a r_soil parametrisation is equally valid

And why couldn't those implementations simply overload the method? Do you mean it would be better to dispatch on an explicit parameterization of the ground evaporation rather than the whole ET type?

src/solvers/solvers.jl
bgroenks96: Technically, yes. And it seems to depend heavily on the drag coefficient and skin conductivity. I have nevertheless kept the fixed-point solver in for now... maybe it will come in handy for testing purposes.

test/surface_hydrology/evapotranspiration_tests.jl
bgroenks96: I think an LLM did that. I'm indifferent to it and also to consistency in its usage.

Project.toml
bgroenks96: Fixed in [4d4993a](https://github.com/NumericalEarth/Terrarium.jl/pull/135/commits/4d4993a4c272e53613b0627530be4f3fc41b0200)

src/solvers/root_solvers.jl
bgroenks96: Done in [53ead2c](https://github.com/NumericalEarth/Terrarium.jl/pull/135/commits/53ead2c8b8696be69a7f9cb67e69dee92a17683d)

src/processes/surface_hydrology/evapotranspiration/canopy_evapotranspiration.jl
bgroenks96: Done in [7af1661](https://github.com/NumericalEarth/Terrarium.jl/pull/135/commits/7af1661efedc66996d0730b21933aee0f0ba715e)

src/processes/surface_hydrology/evapotranspiration/canopy_evapotranspiration.jl
bgroenks96: Fixed in [7af1661](https://github.com/NumericalEarth/Terrarium.jl/pull/135/commits/7af1661efedc66996d0730b21933aee0f0ba715e)

src/processes/surface_hydrology/evapotranspiration/canopy_evapotranspiration.jl
bgroenks96: Fixed in [7af1661](https://github.com/NumericalEarth/Terrarium.jl/pull/135/commits/7af1661efedc66996d0730b21933aee0f0ba715e)

src/processes/surface_hydrology/evapotranspiration/canopy_evapotranspiration.jl
bgroenks96: Fixed in [7af1661](https://github.com/NumericalEarth/Terrarium.jl/pull/135/commits/7af1661efedc66996d0730b21933aee0f0ba715e)

src/processes/surface_energy/surface_energy_balance.jl
bgroenks96: Fixed comment in [7af1661](https://github.com/NumericalEarth/Terrarium.jl/pull/135/commits/7af1661efedc66996d0730b21933aee0f0ba715e)

src/processes/surface_hydrology/evapotranspiration/bare_ground_evaporation.jl
bgroenks96: I think this is out of scope. We probably will not be implementing T_s dependent resistance terms in Terrarium anyway, since I think the current consensus is that more sophisticated surface flux schemes should be implemented in NumericalEarth.jl.

src/processes/surface_hydrology/evapotranspiration/canopy_evapotranspiration.jl
bgroenks96: `1 / r_a` is the aerodynamic conductance. The scaling by `f_can` is what makes this the canopy conductance. So I don't really see the problem here.

src/processes/surface_hydrology/evapotranspiration/canopy_evapotranspiration.jl
bgroenks96: For now, I think we do.

src/processes/surface_hydrology/evapotranspiration/evapotranspiration_base.jl
bgroenks96: Fixed in [106cd69](https://github.com/NumericalEarth/Terrarium.jl/pull/135/commits/106cd69fe4d528971ed929c0bdf8820eb9c71bf3)

src/processes/surface_hydrology/evapotranspiration/bare_ground_evaporation.jl
bgroenks96: I can change this to `g_gnd` but otherwise, same response as for the canopy. I think the scaling factors (if present) are part of the ground conductance; otherwise it's just the aerodynamic conductance which need not be specific to the ground or canopy.

bgroenks96: Renamed in [f27f684](https://github.com/NumericalEarth/Terrarium.jl/pull/135/commits/f27f6840c8e74fd1d8b848a71b33f0d9cdca55eb)

src/processes/surface_hydrology/evapotranspiration/bare_ground_evaporation.jl
olivierbonte: Ahn okay, no that's sound like a good decision to have all flux parametrisations over all types of surfaces (land, ocean...) in NumericalEarth.jl. Then this is comment is indeed out of scope.

src/processes/surface_hydrology/evapotranspiration/bare_ground_evaporation.jl
olivierbonte: For the current state of the code: `ground_evaporation_resistance_factor` returns $\Beta$, while `ground_evaporation_conductance` returns `\Beta/r_a`. I thought it was a bit confusing as one is not the inverse of the other. But I guess that's why you use the `_factor`? 

And you're right for the overloading, didn't think that one through.

src/processes/surface_hydrology/evapotranspiration/bare_ground_evaporation.jl
olivierbonte: Okay that's fine! In the current approach there is indeed only one aerodynamic conductance, so no need to specify on that, that's only relevant once two-source models would come in.

docs/src/processes/surface_energy/skin_temperature.md
olivierbonte: Maybe best to adapt docs to reflect the current standard: solving to the root with a numerical method from Rootsolvers.jl

olivierbonte: Maybe good to add a brief docs section on the solvers e.g. under `docs/src/utils` and link to it?

src/processes/surface_energy/skin_temperature.jl
```diff
@@ -170,12 +190,63 @@ Compute the ground heat flux from the surface net radiation and sensible/latent
...
+        skinT::ImplicitSkinTemperature,
+        seb::AbstractSurfaceEnergyBalance,
+        seb_args...
+    )
+    # Compute all fluxes based on current skin temperature (this includes ground heat flux already)
+    compute_surface_energy_fluxes!(out, i, j, grid, fields, seb, seb_args...)
+    # Compute skin temperature
+    Ts_implicit = compute_skin_temperature(i, j, grid, fields, skinT)
+    Ts_prev = out.skin_temperature[i, j, 1]
+    return Ts_prev - Ts_implicit
```
olivierbonte, line 233: Is there a specific reason why you calculate the residual on T_s?
I think the more classic approach is to calculate the surface energy balance residual: F(T_s) = R_n + H_l + H_S + G = 0

src/solvers/root_solvers.jl
```diff
@@ -0,0 +1,119 @@
+import RootSolvers
```
olivierbonte: General note on the solvers: there is no documentation on this yet. We could that add that here, follow up PR can also work.

olivierbonte: Additional note: maybe better to have this import in Terrarium.jl? That's also how we did if for Thermodynamics.jl

src/solvers/solvers.jl
```diff
@@ -0,0 +1,54 @@
...
+    func::F
+
+    dfunc::DF
```
olivierbonte, line 20: I like the fact that you can choose to provide the derivative of your residual to the target. Would be nice to have a bit more on the docstrings of what is expected here.

src/solvers/root_solvers.jl
```diff
@@ -0,0 +1,119 @@
...
+
+    function dresidual!(x)
+        target_field[indices...] = x
+        return dstep_func!(out, indices..., grid, fields, func_args...; func_kwargs...)
+    end
+
+    return x -> (residual!(x), dresidual!(x))
+end
+
+@propagate_inbounds function build_residual(
```
olivierbonte, line 95: Ideally some docstrings: default = Newton -> 2 options: 1) provide derivative function in ObjectiveFunction 2) fallback to forward finite differences.

Just out of interest: is there a specific reason why you chose not to use FiniteDiff.jl or FiniteDifferences.jl for this fallback?

test/surface_hydrology/evapotranspiration_tests.jl
```diff
@@ -0,0 +1,282 @@
...
+# Surface humidity flux partitioning (canopy)
+# ==============================================================================
+
+@testset "surface_humidity_flux canopy" begin
+    # Total flux is the sum of the partitioned components
+    E_gnd = 1.0e-6
+    E_can = 2.0e-6
+    T_can = 3.0e-6
+    @test E_gnd + E_can + T_can ≈ 6.0e-6
```
olivierbonte: This doesn't test anything right?

test/surface_hydrology/evapotranspiration_tests.jl
```diff
@@ -0,0 +1,282 @@
...
+    g_stm = 0.1
+    f_can = 0.8
+    β = 0.7
+    Δq_skin = 0.01    # humidity difference at skin temperature
+    Δq_ground = 0.005 # humidity difference at ground temperature
+
+    # Step 1: conductances
+    g_trp = transpiration_conductance(canopy_ET, rₐ, g_stm)
+    g_can = canopy_evaporation_conductance(canopy_ET, f_can, rₐ)
+    g_gnd = β / rₐ
```
olivierbonte: for consistency with other conductances: use ground_evaporation_conductance here?

olivierbonte (commented):
The updates look good, thanks for the changes! Only real remaining questions for me are whether or not to include some docs for the solvers in this PR + residual on energy balance or or on Ts.

src/processes/surface_energy/skin_temperature.jl
bgroenks96: I guess you mean with $G$ computed directly as the temperature gradient right? This is mostly an artifact of the previous fixed-point approach. It also makes it easier to keep `FixedPointSolver` as an option (for now).

But mathematically this is equivalent I think. It's just a different arrangement of terms.

src/solvers/root_solvers.jl
bgroenks96: Yeah I initially did it with `FiniteDifferences` and it was slower with type stability issues. Decided to just keep it simple (and fast) for now.

test/surface_hydrology/evapotranspiration_tests.jl
bgroenks96: Hmm.. yeah, ok I need to go through this test set and clear out the slop.

src/processes/surface_hydrology/evapotranspiration/bare_ground_evaporation.jl
bgroenks96: Oh, I see your point. Yes that is a bit unfortunate, but that is indeed the point of adding `_factor`. Maybe we can rename this in the future if we can think of better names.

src/solvers/root_solvers.jl
olivierbonte: Alright that makes sense, thanks for the clarification!

src/processes/surface_energy/skin_temperature.jl
olivierbonte: It is indeed the same equation you are solving for, only difference:
- this approach: T_s -> H, LE and R_n -> G as residual -> T_s_new -> residual you want to set to zero is T_s - T_s_new
- other approach: T_S -> H, LE, R_n and G(= k DeltaT/DeltaZ) = residual in W/m^2 that is retruned

second approach has the advantage that you can set energy balance residual as the thing you want to get under some specific tolerance (e.g. 1 W/m^2). 

But if you prefer to keep the T_s residual around for now that is also okay!

test/surface_hydrology/evapotranspiration_tests.jl
bgroenks96: Fixed in [8ac055a](https://github.com/NumericalEarth/Terrarium.jl/pull/135/commits/8ac055a6ac80200b179140790b3fb249dbd341ea)

src/solvers/solvers.jl
bgroenks96: Done in [66e2c10](https://github.com/NumericalEarth/Terrarium.jl/pull/135/commits/66e2c108672a6f0c77d7cb6a59b4dbfd3d06bf25)

test/surface_hydrology/evapotranspiration_tests.jl
bgroenks96: Fixed in [8ac055a](https://github.com/NumericalEarth/Terrarium.jl/pull/135/commits/8ac055a6ac80200b179140790b3fb249dbd341ea)

src/solvers/root_solvers.jl
bgroenks96: Fixed in [1d2c0c1](https://github.com/NumericalEarth/Terrarium.jl/pull/135/commits/1d2c0c1f8a600e20db21f79c289696e22bdfbfb7)

docs/src/processes/surface_energy/skin_temperature.md
bgroenks96: Done in [e4b9388](https://github.com/NumericalEarth/Terrarium.jl/pull/135/commits/e4b9388c986658e18e63d6888ea978bc1b09c200) and [f6e68a4](https://github.com/NumericalEarth/Terrarium.jl/pull/135/commits/f6e68a48740d83364729efb16c1a7febf8930ba4)

docs/src/solvers/solvers.md
```diff
@@ -0,0 +1,54 @@
...
+[RootSolvers.jl](https://github.com/CliMA/RootSolvers.jl) (e.g. Newton's method). When the [`ObjectiveFunction`](@ref) carries no analytic derivative, a finite-difference derivative is formed automatically so that derivative-based methods such as Newton's method can still be used.
+This is the default solver for the implicit skin temperature, configured by [`default_skin_temperature_solver`](@ref).
+
+```@docs; canonical = false
+RootSolver
+```
+
+## Fixed-point solvers
+
+[`FixedPointSolver`](@ref) implements a simple Picard iteration. Because the objective returns
```
olivierbonte: Maybe add  `x = g(x)` after 'simple Picard Iteration'

docs/src/solvers/solvers.md
```diff
@@ -0,0 +1,54 @@
...
+the residual `F(x) = x - g(x)`, the updated iterate is recovered as `g(x) = x - F(x)`. Updates
```
olivierbonte: The residual doesn't have to be formulated in this way, I think it is more clear to only state that the update iterate is approximated as `g(x) = x - F(x)`

olivierbonte (commented):
Looking great! Thanks for the extra documentation, your last commit also fixes the tests (when testing locally) :)

docs/src/solvers/solvers.md
bgroenks96: Fixed [17f6940](https://github.com/NumericalEarth/Terrarium.jl/pull/135/commits/17f694018c50f10746a04fecbd1dd0c98c5bbfd2)

docs/src/solvers/solvers.md
bgroenks96: Fixed [17f6940](https://github.com/NumericalEarth/Terrarium.jl/pull/135/commits/17f694018c50f10746a04fecbd1dd0c98c5bbfd2)

olivierbonte (approved):
Thanks for changing this description! All approved

PR #136, opened 2026-06-29
Test namespaced integration

maximilian-gelbrecht:
Related to #125, but stand-alone, this adds a test that integrating a model with variables in namespaces actually works and also fixes a minor issue that prevented it from working correctly before

### DIFF
--- a/src/state_variables.jl
+++ b/src/state_variables.jl
@@ -382,7 +382,7 @@ function initialize(
     namespaces = map(vars.namespaces) do ns
         ns_bcs = get(boundary_conditions, varname(ns), (;))
         ns_fields = get(fields, varname(ns), (;))
-        initialize(ns.vars, grid; clock, boundary_conditions = ns_bcs, fields = ns_fields)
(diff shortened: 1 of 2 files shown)


src/state_variables.jl
```diff
@@ -382,7 +382,7 @@ function initialize(
     namespaces = map(vars.namespaces) do ns
         ns_bcs = get(boundary_conditions, varname(ns), (;))
         ns_fields = get(fields, varname(ns), (;))
-        initialize(ns.vars, grid; clock, boundary_conditions = ns_bcs, fields = ns_fields)
+        initialize(ns.vars, grid; clock, timestepper, boundary_conditions = ns_bcs, fields = ns_fields)
```
bgroenks96, line 385: Is this the current API or the updated API from #125? I guess the current one, right?

bgroenks96 (approved):

src/state_variables.jl
maximilian-gelbrecht: Yes the current one

PR #138, opened 2026-07-03
Reactant compatibility 

maximilian-gelbrecht:
This is a mostly vibe coded and guided attempt to run Terrarium with Reactant. 

It runs correctly for some steps already which is really great. So far this is only tested on CPU though. 

Some unexpected changes that needed to be made: 

* Within the compile context we can't `@assert`, so some of those were moved
* There seems to be an Oceananigans issue when the vertical coordinates are handed over as arrays and not as ranges. I hope that I can come up with a fix over at Oceananigans.jl for this quickly. Once that's done, we should be able to revert the changes made here to the vertical discretisation. 

As expected (kind of) model initialisation is done on regular CPU and then everything is transferred to Reactant. This isn't ideal, and we might revise this at a later point, but it's the quickest way to make it work. 

At the current state this is just an early draft, that also includes a bunch claude-comments that I'll remove later.


ext/TerrariumReactantExt/integrator.jl
```diff
@@ -0,0 +1,50 @@
...
+#
+# `timestep!`/`run!` on a device integrator compile a single step with Reactant (once, cached) and
+# drive it from a host loop. The compiled core calls the timestepper-level `timestep!`
+# (`timestep!(integrator, timestepper, Δt)`), which is NOT overridden here — this is what avoids
+# recursively re-entering the compiling wrapper.
+
+const ReactantIntegrator = ModelIntegrator{<:Any, <:RARCH}
+
+# Compiled steps keyed by (integrator identity, Δt). Δt is a trace constant, so a new Δt recompiles.
+const COMPILED_STEPS = Dict{Any, Any}()
```
maximilian-gelbrecht: This will probably be changed. I am don't like this approach so much.

bgroenks96:
> There seems to be an Oceananigans issue when the vertical coordinates are handed over as arrays and not as ranges. 

I don't understand. The vertical coordinates should be arrays? Oceananigans also allows them to be dynamic (sigma coordinates).

bgroenks96:
Also, I would like to start keeping track of LLM-generated (and possibly human-augmented) implementation plans as developer documentation. So please move your plan files into a new folder `docs/dev/reactant` and maybe prefix them with a number (or date?) like `1_reactant_integration_plan.md` so the file names have a natural order roughly aligned with the revision history.

I think all plan documents should also include the following as metadata:
- The original prompt
- Any subsequent revisions to the instructions in the original prompt
- Status of the implementation (planned vs. implemented; Claude usually does this anyway)
- Date and git commit SHA1 on which the plan was first drafted
- Date and git commit SHA1 on which the (initial) implementation was completed

I will also add this to the `AGENTS.md`.

maximilian-gelbrecht:
> > There seems to be an Oceananigans issue when the vertical coordinates are handed over as arrays and not as ranges.
> 
> I don't understand. The vertical coordinates should be arrays? Oceananigans also allows them to be dynamic (sigma coordinates).

Yes, but exactly this currently doesn't work with Reactant. But the PR is already open, so that'll be fixed quickly.

maximilian-gelbrecht:
> Also, I would like to start keeping track of LLM-generated (and possibly human-augmented) implementation plans as developer documentation. So please move your plan files into a new folder `docs/dev/reactant` and maybe prefix them with a number (or date?) like `1_reactant_integration_plan.md` so the file names have a natural order roughly aligned with the revision history.
> 
> I think all plan documents should also include the following as metadata:
> 
> * The original prompt
> * Any subsequent revisions to the instructions in the original prompt
> * Status of the implementation (planned vs. implemented; Claude usually does this anyway)
> * Date and git commit SHA1 on which the plan was first drafted
> * Date and git commit SHA1 on which the (initial) implementation was completed
> 
> I will also add this to the `AGENTS.md`.

I am not sure if it's really beneficial to keep them permanently in the repo. Committing them while working on the feature makes sense both for working on it and for reviewing, but why would you keep them after everything is done in the repo? I personally keep them in my notes app though.

bgroenks96:
It documents design decisions and the development process, also making it at least partially reproducible.

Ideally, these kinds of development notes should be written and documented even without LLMs, but most of us (or at least I) don't have the time or patience to do it. LLMs just make them (almost) free to add. Seems like a win-win to me.

And sure, you could store them somewhere else, but then they aren't easily accessible to all collaborators.

ext/TerrariumReactantExt/integrator.jl
maximilian-gelbrecht: Revised

maximilian-gelbrecht:
In practice it's kind of hard to keep track of the interactivity though. I don't just have a single prompt. I guide the LLM into the right direction during its implementation. Over multiple sessions these plans are not really nice to read anymore and also aren't really that great anymore for reproduction. I don't mind saving them somewhere, but I just don't find them that useful.

maximilian-gelbrecht:
To the actual topic here: 

* I went through one more revision of the core implementation (the extension) now, and I am reasonably happy with it
* Currently the `soil_heat_coumn.jl` and `soil_heat_global.jl` examples are used with Reactant, which other examples should we add? 
* I also added tests and example to differentiate the models with Reactant 
* I work on an example to set up a simple NN-based model with Reactant and Lux, currently half finished. 

I'd wait until https://github.com/CliMA/Oceananigans.jl/pull/5769 is merged and released and then do one more round over this to clean up everything. But this is already in a decent state now here. A bit to-do left is to try everything on GPU as well, but I don't expect major hurdles there.

maximilian-gelbrecht:
@bgroenks96 If you have some time, I wouldn't mind a review. 

This is in a state that I would regard as almost ready. I don't think we need full Reactant compatibility of every single model in one PR. This here is about the first models and the basic infrastructure for Reactant. 

* This PR just adapts the models from the `soil_heat_global.jl` and `soil_heat_column.jl` examples. If we should add more, let me know! 
* All changes related to the z-dimension are subject to change once the PR over at Oceanananigs is merged and released
* One Terrarium feature that isn't adapted here and that will be quite tough to adapt are inputs from `FieldTimeSeries` but I also have to see if Oceananigans already has done some work for that. The temporal interpolation in particular is unlikely to work under Reactant tracing. There are just too many runtime conditionals in there. Turning just into a time series that can be indexed directly will probably work though.

maximilian-gelbrecht:
@bgroenks96 I went over this once more, and also the Oceananigans issue was fixed. It's currently not released yet so this depends on `main`, but this could be merged from my side once the next Oceananigans version is released. The latest revision and the fix simplified the whole approach even more.

examples/hybrid_models/neural_snow_melt.jl
```diff
@@ -0,0 +1,419 @@
...
+# Inputs and targets are normalized; the normalization
+# constants are stored in the `NeuralSnowMelt` process and undone at inference.
+
+const NF = Float32
+reference = DegreeDaySnow(NF)
+
+T_train = reshape(collect(range(NF(-30), NF(30), length = 512)), 1, :)   # (1, N) row of temperatures
+M_train = melt.(Ref(reference), T_train)                                  # (1, N) target melt rates
+
+T_mean = mean(T_train)
```
bgroenks96, line 195: Not that it really matters, but why not just use physical scaling? i.e. convert to Kelvin and then divide by 273.15?

I think for the melt rate I would also just consider using a physical rather than "data"-driven scaling, maybe just convert to mm?

examples/hybrid_models/neural_snow_melt.jl
```diff
@@ -0,0 +1,419 @@
...
+using Terrarium
+using Reactant, CUDA          # CUDA is required by Reactant's kernel integration, even on CPU
+using Lux, Enzyme, Optimisers, Random
+using KernelLux               # device-safe `apply_in_kernel` / `kernelize` for Lux in kernels
+using StaticArrays: SA
+using Statistics: mean, std
+using KernelAbstractions: @index, @kernel
+using Terrarium: launch!, XY, interior, AbstractProcess
+using Terrarium: Adapt          # Adapt is re-exported by Terrarium (transitive dep)
+import Lux.MLDataDevices as MLDD
```
bgroenks96, line 31: That's a weird choice of submodule name from the Lux authors...

examples/hybrid_models/neural_snow_melt.jl
```diff
@@ -0,0 +1,419 @@
...
+#
+# We build a small MLP, move it (and the data) to the Reactant device, and train with
+# `Lux.Training.single_train_step!` using the Enzyme AD backend. On the Reactant device this
+# compiles the train step with XLA; each step computes the loss and its gradient with Enzyme and
+# updates the parameters with Adam.
+
+const H = 16
+mlp = Lux.Chain(Lux.Dense(1 => H, tanh), Lux.Dense(H => 1))
+
+rdev = MLDD.reactant_device()
```
bgroenks96, line 211: normally I like explicit namespace references, but here with the `MLDD` alias it's a bit confusing unless you were paying close attention to the imports (which probably no one does). Maybe this is a case where it's better to `using Lux.MLDataDevices: reactant_device` since the name `reactant_device()` on its own is fairly self-explanatory.

examples/hybrid_models/neural_snow_melt.jl
```diff
@@ -0,0 +1,419 @@
...
+rng = Random.Xoshiro(0)
+ps, st = Lux.setup(rng, mlp) |> rdev
+data = (rdev(T_norm), rdev(M_norm))
```
bgroenks96: This is of course obvious to everyone who is used to machine learning frameworks but maybe it's worth adding an explicit comment here explaining that `rdev(T_norm)` is the thing actually moving the data onto the device?

examples/hybrid_models/neural_snow_melt.jl
```diff
@@ -0,0 +1,419 @@
...
+end
+
+snow_ddm = run_snow(reference)
+snow_nn = run_snow(neural)
+
+snow_diff = maximum(abs, snow_ddm .- snow_nn)
+println("Ran both snow models for 30 days on $(length(_lats)) columns.")
+println("Max |S_ddm − S_nn| after offline training: ", snow_diff, " m  (mean snow ", mean(snow_ddm), " m)")
+
+# ## Online finetuning: training *through* the full Terrarium model
```
bgroenks96:
```suggestion
# ## Online fine-tuning: training *through* the full Terrarium model
```

Sorry, pet peeve of mine... I don't like the compound form of fine-tuning.

examples/hybrid_models/neural_snow_melt.jl
```diff
@@ -0,0 +1,419 @@
...
+# rollout to the network weights. This is the differentiable-simulation idea, and it is what
+# Reactant + Enzyme make possible.
+#
+# It can help even in this toy setting: the offline objective fits `M(T)` pointwise, whereas the
+# online objective matches the simulated snow trajectory directly — the quantity we care about —
+# so a small pointwise melt error that would accumulate over the rollout gets corrected.
+#
+# ### Two ways to evaluate the network — same result
+#
+# We run the online finetuning **twice**, with the two processes defined above, to show to both them
```
bgroenks96:
```suggestion
# We run the online fine-tuning **twice**, with the two processes defined above, to show to both them
```

examples/hybrid_models/neural_snow_melt.jl
```diff
@@ -0,0 +1,419 @@
...
+
+snow_diff = maximum(abs, snow_ddm .- snow_nn)
+println("Ran both snow models for 30 days on $(length(_lats)) columns.")
+println("Max |S_ddm − S_nn| after offline training: ", snow_diff, " m  (mean snow ", mean(snow_ddm), " m)")
+
+# ## Online finetuning: training *through* the full Terrarium model
+#
+# The training above was *offline*: we fit the melt law `M(T)` pointwise, independently of the
+# simulation. *Online* training instead differentiates through the model's own time integration.
+# Here we do this with the **full Terrarium model** — a `SnowModel` containing the neural process,
```
bgroenks96:
```suggestion
# Here we do this with the full `SnowModel`, which contains the neural network,
```

I would also avoid the term "neural process" since this actually has a specific meaning.

examples/hybrid_models/neural_snow_melt.jl
```diff
@@ -0,0 +1,419 @@
...
+# stepped by `run_timesteps!` — on the Reactant device: we roll the model forward, compare the
```
bgroenks96:
```suggestion
# on the Reactant device: we roll the model forward, compare the
```

examples/hybrid_models/neural_snow_melt.jl
bgroenks96: Maybe just CTRL-F replace the rest...

examples/hybrid_models/neural_snow_melt.jl
```diff
@@ -0,0 +1,419 @@
...
+
+function snow_loss(integrator, S0, S_ref, Nt)
+    copyto!(interior(integrator.state.snow_storage), S0)     # reset the rollout
+    run_timesteps!(integrator, Δt_ft, Nt)                    # step the full Terrarium model
+    return mean(abs2, interior(integrator.state.snow_storage) .- S_ref)
+end
+function snow_grad!(integrator, dintegrator, S0, S_ref, Nt)
+    _, loss = Enzyme.autodiff(
+        Enzyme.set_runtime_activity(Enzyme.ReverseWithPrimal), snow_loss, Enzyme.Active,
+        Enzyme.Duplicated(integrator, dintegrator),
```
bgroenks96, line 352: As an aside, I would like to see if we can add some APIs to automate some of this. I don't think the user should need to write `grad` functions. Also, I think we should see if there's a way to not always `Duplicate` the whole `integrator`... that kind of defeats the purpose of separating model and state.

examples/hybrid_models/neural_snow_melt.jl
```diff
@@ -0,0 +1,419 @@
...
+end
+
+# Adam update of the network parameters. Because rolling the melt through `Nt·Δt ≈ 10⁶` seconds
+# amplifies its gradient enormously, an adaptive optimizer (which rescales by the gradient
+# magnitude) is far more robust here than plain gradient descent. We optimize the Lux parameters
+# `ps` directly; the optimizer runs eagerly on the host over the device arrays — which are shared
+# with `integrator.model.snow_melt.ps`, so the model updates in place. (Eager, not traced, so
+# Adam's scalar state is a plain number.) The compiled gradient function is passed in so the same
+# loop works for either process.
+function finetune!(compiled_grad!, integrator, S0, S_ref; epochs = 100, lr = NF(1.0e-3), label = "")
```
bgroenks96, line 365: this is fine because Julia conventions

examples/hybrid_models/neural_snow_melt.jl
```diff
@@ -0,0 +1,419 @@
...
+end
+
+# Each approach starts from an independent copy of the offline-trained weights (a device→host→device
+# round trip), so finetuning one does not disturb the other. The in-kernel CPU process uses
+# `kernelize`d parameters; the batched CPU process uses plain arrays (ordinary `Lux.apply`).
+fresh_offline_ps() = rdev(cdev(tstate.parameters))
+
+snow_nn_ft_kernel = finetune_and_run(
+    NeuralSnowMelt(mlp, fresh_offline_ps(), tstate.states, T_mean, T_std, M_scale),
+    (ps, st) -> neural_process(mlp, ps, st);
```
bgroenks96: would also suggest getting rid of the `neural_process` naming here and above

ext/TerrariumReactantExt/integrator.jl
```diff
@@ -0,0 +1,67 @@
...
+# Reactant specialization of `Terrarium.run_timesteps!` (generic host loop defined in src). Here
+# `Reactant.@trace` compiles the loop to a single `stablehlo.while` (rather than unrolling `Nt`
+# copies of the step into the trace); this is the function compiled by `run!` and mirrors
+# Oceananigans' `run_timesteps!` test helper.
+#
+# `checkpointing` is forwarded to `@trace` and controls reverse-mode-AD memory: `false` (default)
+# stores every intermediate state; a scheme such as `Reactant.Periodic(n)`/`Reactant.Binomial(n)`
+# stores only `n` checkpoints and recomputes the rest. It has no effect on a pure forward run but
+# matters when this function is differentiated (see the autodiff example).
+function Terrarium.run_timesteps!(integrator::ReactantIntegrator, Δt, Nt, checkpointing = false)
```
bgroenks96, line 20: Nice, I love that we can use dispatch for this!

ext/TerrariumReactantExt/transfer.jl
```diff
@@ -0,0 +1,31 @@
...
+        clock = nothing,
+        inputs = Terrarium.InputSources(NF),
+        boundary_conditions = (;),
+        initializers = (;),
+        fields = (;),
+    ) where {NF}
+    # Traced device clock (unless the caller supplied one).
+    device_clock = isnothing(clock) ?
+        Oceananigans.TimeSteppers.Clock(get_field_grid(get_grid(model))) : clock
+    return @invoke Terrarium.initialize(
```
bgroenks96: wait, sorry, what is `@invoke`?

src/processes/soil/energy/soil_energy_closures.jl
```diff
@@ -133,33 +133,31 @@ end
...
-        one(sat)
-    else
-        # Case 2a: -Lθ ≤ U ≤ 0 -> phase change
-        # Case 2b: U < -Lθ -> frozen (zero)
-        (U >= -Lθ) * (one(sat) - safediv(U, -Lθ))
-    end
+    # Case 1: U ≥ 0 -> thawed (liq = 1)
+    # Case 2a: -Lθ ≤ U < 0 -> phase change (liq = 1 - U/(-Lθ))
+    # Case 2b: U < -Lθ -> frozen (liq = 0), enforced by the (U ≥ -Lθ) factor.
+    # Result is identical to the previous `if/else`: `ifelse` merely also evaluates the
```
bgroenks96: Remove this comment

src/processes/soil/stratigraphy/soil_composition.jl
```diff
@@ -8,28 +8,43 @@ and a mixture of organic and mineral solid material.
...
 
     "Liquid (unfrozen) fraction of pore water"
-    liquid::NF = 1.0
+    liquid::NF
 
     "Parameterization of the solid phase (matrix) of the soil"
-    solid::Solid = MineralOrganic(texture = SoilTexture(), organic = 0.0)
+    solid::Solid
 
-    # Scalar constructor
```
bgroenks96, line 24: Oops, this was leftover from some old code...

src/processes/soil/stratigraphy/soil_volume.jl
```diff
@@ -8,28 +8,43 @@ and a mixture of organic and mineral solid material.
...
     "Liquid (unfrozen) fraction of pore water"
-    liquid::NF = 1.0
+    liquid::NF
 
     "Parameterization of the solid phase (matrix) of the soil"
-    solid::Solid = MineralOrganic(texture = SoilTexture(), organic = 0.0)
+    solid::Solid
 
-    # Scalar constructor
+    # Positional constructor
```
bgroenks96: this can be removed though

src/processes/soil/stratigraphy/soil_composition.jl
```diff
@@ -8,28 +8,43 @@ and a mixture of organic and mineral solid material.
...
+    # Positional constructor
     function SoilVolume(porosity::NF, saturation::NF, liquid::NF, solid::AbstractSoilMatrix{NF}) where {NF <: Number}
-        @assert zero(NF) <= porosity <= one(NF)
```
bgroenks96, line 26: I don't love that we can't use `@assert`s now... is there really no workaround for this?

src/processes/soil/stratigraphy/soil_volume.jl
```diff
@@ -8,28 +8,43 @@ and a mixture of organic and mineral solid material.
...
     end
 end
 
+"""
+    $SIGNATURES
+
+Validating keyword constructor for `SoilVolume`; enforces that `porosity`, `saturation`,
+and `liquid` lie in the unit interval.
+"""
+function SoilVolume(;
```
bgroenks96: Ok... I guess that's kind of a workaround...

src/processes/soil/stratigraphy/soil_composition.jl
```diff
@@ -76,19 +91,30 @@ Soil matrix consisting of a simple, homogeneous mixture of mineral and organic m
 Properties:
 $TYPEDFIELDS
 """
-@kwdef struct MineralOrganic{NF} <: AbstractSoilMatrix{NF}
+struct MineralOrganic{NF} <: AbstractSoilMatrix{NF}
```
bgroenks96, line 93: Are we officially retiring `@kwdef` constructors now? Not opposed to that, but just want to clarify.

src/processes/surface_hydrology/canopy_interception/canopy_interception.jl
```diff
@@ -89,7 +89,7 @@ Compute the canopy saturation fraction as `w_can / w_can_max`.
 @inline function compute_canopy_saturation_fraction(canopy_interception::PALADYNCanopyInterception{NF}, w_can, LAI, SAI) where {NF}
     # Compute the wet canopy fraction
     w_can_max = canopy_interception.w_can_max * (LAI + SAI)
-    f_can = w_can_max > 0 ? w_can / w_can_max : zero(NF)
+    f_can = ifelse(w_can_max > zero(NF), w_can / w_can_max, zero(NF))
```
bgroenks96: I think if you use `ifelse` then you probably need to use `safediv`, no? Otherwise you can end up with `NaN` gradients.

src/utils/kernel_utils.jl
```diff
@@ -8,11 +8,12 @@ This implementation performs a linear scan over the z-axis and thus has time com
...
     idx = -1
     for k in 1:n
-        if idx < 0 && condition_func(field[i, j, k])
-            idx = k
-        end
+        found = (idx < 0) & condition_func(field[i, j, k])
+        idx = ifelse(found, k, idx)
     end
-    return idx > 0 ? z_nodes[idx] : z_nodes[n]
+    # Select the index (not the loads): ifelse evaluates both branches, so indexing in the
```
bgroenks96: should this be "not the nodes"...>

test/reactant/correctness.jl
```diff
@@ -0,0 +1,166 @@
+# Generic CPU-vs-Reactant correctness machinery.
+#
+# The design mirrors `SpeedyWeather/test/reactant/`: build the *same* model on two
+# architectures, sync one state onto the other so both start bit-identical, advance both,
```
bgroenks96:
```suggestion
# architectures, sync one state onto the other so both start identical, then advance both
```

I don't understand Claude's obsession with adding "bit" everywhere.... bit-for-bit, bit-identical, etc.

test/reactant/correctness.jl
```diff
@@ -0,0 +1,166 @@
...
+
+_cpu_array(f::AbstractField) = Array(interior(f))
+_cpu_array(a::AbstractArray) = Array(a)
+_cpu_array(::Any) = nothing          # skip non-array auxiliary entries (e.g. scalar refs)
+
+# --- state traversal --------------------------------------------------------------------
+
+# Collect all comparable field arrays (prognostic + auxiliary), recursing into namespaces,
+# keyed by a dotted path so names stay unambiguous across namespaces.
+function collect_state_arrays(state, prefix = "")
```
bgroenks96, line 32: Why is all of this necessary? Can't you just use `getfields` and then extract the arrays from them?

test/reactant/Project.toml
```diff
@@ -0,0 +1,16 @@
...
+Reactant = "3c362404-f566-11ee-1572-e11a4b42c853"
+RingGrids = "d1845624-ad4f-453b-8ff4-a8db365bf3a7"
+Statistics = "10745b16-79ce-11e8-11f9-7d13ad32a3b2"
+Terrarium = "80418d68-07fa-499d-ae2b-c12e531f5cd8"
+Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
+
+[sources]
+Oceananigans = {url = "https://github.com/CliMA/Oceananigans.jl.git", rev = "main"}
+
+[compat]
```
bgroenks96, line 11: Julia 1.12 compat? or no?

test/reactant/runtests.jl
```diff
@@ -0,0 +1,45 @@
...
+            "julia --project=test/reactant test/reactant/runtests.jl"
+    )
+end
+
+using Terrarium
+using Reactant
+using CUDA   # required by Reactant's KernelAbstractions integration, even on CPU
+using Test
+
+# Tolerances: XLA reorders floating-point ops, so exact equality is not expected. Start loose
```
bgroenks96: Rewrite this comment, it's way to Claude-ish... using imprecise language like "far quieter" whatever that means.

test/reactant/setup.jl
```diff
@@ -0,0 +1,65 @@
...
+    bcs = PrescribedSurfaceTemperature(:T_ub, NF(1))
+    inits = (temperature = (x, z) -> NF(-1) - NF(0.05) * z,)
+    return (; model, boundary_conditions = bcs, initializers = inits, Δt = NF(600))
+end
+
+# --- :soil_heat_global — global ColumnRingGrid SoilModel (heat conduction) --------------
+# Mirrors examples/simulations/soil_heat_global.jl, but with a synthetic all-land mask
+# (no NetCDF inputs in CI) and a purely functional, smooth surface BC (no array gather).
+
+function build_model(::Val{:soil_heat_global}, arch, NF)
```
bgroenks96, line 50: what's with the `Val` dispatch pattern here?

bgroenks96 (changes requested):
Looking good! I am still somehow amazed that the NN example actually works. That's a big step!

examples/hybrid_models/neural_snow_melt.jl
maximilian-gelbrecht: A physical scaling only would not be sufficient. You need normalised inputs for the NN with 0 mean and 1 std. 

Doing a physical scaling on top of the normalisation would be something that one could still do, but I would probably not make this simple demonstration more complicated then it needs to be.

examples/hybrid_models/neural_snow_melt.jl
maximilian-gelbrecht: Agreed.

examples/hybrid_models/neural_snow_melt.jl
maximilian-gelbrecht: But `reactant_device` could also easily be the name of one of our (or Oceanananigs) device/architecture handling functions. I would therefore prefer the explicit reference here.

test/reactant/correctness.jl
maximilian-gelbrecht: I think in this case it was actually just taking it from my prompt in which I demanded it checks that things are indeed bit-identical.

examples/hybrid_models/neural_snow_melt.jl
bgroenks96: Why do you need standardized inputs? It's conventional, sure, but since when do NNs *need* it?

test/reactant/Project.toml
maximilian-gelbrecht: Why? You seem to have problems with the precompilation under Julia 1.10 but in principle this should work with 1.10 as well.

src/processes/soil/stratigraphy/soil_composition.jl
maximilian-gelbrecht: Just to clarify: we can have `@asserts` in many situation. The situation in which we can't is when we (re-)construct within kernels. That's not something we do that often, right?

ext/TerrariumReactantExt/transfer.jl
maximilian-gelbrecht: `@invoke` is evil dispatch-bypassing logic. It allows to explicitly call a specific method. 

In a way it's hacky solution to have an inheritance in the style of class constructors in Python. 
So, it's hacky, but the alternative would to just repeat the complete code of the `initialize` from the main module. That would also be bad from the point of code repetition and maintainability, so I am fine with the `@invoke` here.

src/processes/soil/stratigraphy/soil_volume.jl
maximilian-gelbrecht: Yes, see comment above

ext/TerrariumReactantExt/integrator.jl
maximilian-gelbrecht: Even though the version in `main` of this function is currently unused. We could in principle use it in `run!` though.

examples/hybrid_models/neural_snow_melt.jl
maximilian-gelbrecht: Okay, they don't strictly need it every single time, but it's just good practice especially in an example that should serve as a general how-to. Generally I think one should always assume they need standardised inputs and then in some situation one can make exceptions. 

Here we use `tanh` as an activation, and that is indeed performing very poorly for non-standardised inputs.

examples/hybrid_models/neural_snow_melt.jl
maximilian-gelbrecht: Yes, agreed. 

I think this example here was just done something very quickly, but it's already working which is great. 

I would say in follow ups we can thing about how we can provide a bit more of interface/framework for this. It will also depend a bit on the use case. Here the training is a kind of specialised emulator-style training. We should probably set up another example that actually trains on data and those setups are easier to transfer and reuse, and then we can hopefully define some interface for that.

src/processes/soil/stratigraphy/soil_composition.jl
maximilian-gelbrecht: A bit similar to above, the main problem (from Reactant side) is when it's used within kernels.

maximilian-gelbrecht: A bit similar to above, the main problem is when it's used within kernels.

src/processes/surface_hydrology/canopy_interception/canopy_interception.jl
maximilian-gelbrecht: Why do you think so? Both branches are computed but only the true on is taken. 

More generally in terms of differentiability, how does `f_can` look like? I am not sure if it would a problem.

test/reactant/setup.jl
maximilian-gelbrecht: This is something Claude came up with. 

I kind of like it because it allows a very compact definition of different model setups. How would you do it instead?

examples/hybrid_models/neural_snow_melt.jl
bgroenks96: The normalization strategy is arbitrary. You can choose physical scales that keep the inputs close to zero and thus should  work with `tanh` activation and most typical weight initializations.

The reason I am bringing this up is because I think the usual practice of standardizing input data often does not make sense in physical systems; e.g. what should the mean and standard deviation of temperature actually be? Obviously, it depends on context.

Ideally we should aim for inductive biases that encourage learning physical relationships, which should probably include keeping physically meaningful scales intact wherever possible.

maximilian-gelbrecht:
@bgroenks96 A new Oceananigans version was released with the patch included, so this could be merged now in principle

test/reactant/correctness.jl
maximilian-gelbrecht: I don't think this would be shorter, our functions for this in the main module don't work across different architectures. I also don't know if we really would want to adjust them for that because it's not needed outside of these correctness tests here.

examples/hybrid_models/neural_snow_melt.jl
maximilian-gelbrecht: In a simple how-to/example just using the normal standardisation is preferable because it works sufficiently well everywhere. 

In a more proper hybrid model we can think about more tailored solutions but I am also really not that convinced that it's particularly important to be honest.

examples/hybrid_models/Project.toml
```diff
@@ -0,0 +1,19 @@
...
+Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
+Reactant = "3c362404-f566-11ee-1572-e11a4b42c853"
+RingGrids = "d1845624-ad4f-453b-8ff4-a8db365bf3a7"
+StaticArrays = "90137ffa-7385-5640-81b9-e52037218182"
+Statistics = "10745b16-79ce-11e8-11f9-7d13ad32a3b2"
+Terrarium = "80418d68-07fa-499d-ae2b-c12e531f5cd8"
+cuDNN = "02a925ec-e4fe-4b08-9a7e-0d78e3d38ccd"
+
+[sources]
+KernelLux = {url = "https://github.com/maximilian-gelbrecht/KernelLux.jl"}
```
bgroenks96, line 18: Are you going to register this?

test/reactant/runtests.jl
```diff
@@ -0,0 +1,44 @@
+# CPU-vs-Reactant correctness tests for Terrarium.
+#
+# Run from the repo root with the dedicated environment (Reactant is NOT a dependency of the
+# main test suite):
+#
+#     julia --project=test/reactant test/reactant/runtests.jl
+#
+# See PLAN_reactant.md for the overall design.
```
bgroenks96: Remove this comment and check the rest of it for Claude-isms.

bgroenks96 (changes requested):

test/reactant/setup.jl
bgroenks96: I guess it's fine.

test/reactant/Project.toml
bgroenks96: It's not *me*, it's an issue with 1.10.11...

https://github.com/EnzymeAD/Reactant.jl/issues/2764

As long as it *should* work then fine to not have a compat entry.

ext/TerrariumReactantExt/transfer.jl
bgroenks96: That is definitely not good and indicates that we need to rethink the dispatch pattern for `initialize`. Fine to keep it for now, but it is not a long-term acceptable solution.

src/processes/soil/stratigraphy/soil_composition.jl
bgroenks96: I guess not, but it was nice to be able to put `@assert` within kernels because they are a bitch to debug otherwise.

src/processes/surface_hydrology/canopy_interception/canopy_interception.jl
bgroenks96: Because I am fairly sure that I ran into this case before with Enzyme where the gradient was `NaN` because one of the branches of `NaN`/`Inf`, and I fixed it with `safediv`. I can't remember the specifics though.

bgroenks96: `f_can` is a linear function of the stored water content, as you can see here. So it should be differentiable except at the thresholds. This should probably be smoothed with a logistic function or something.

examples/hybrid_models/Project.toml
maximilian-gelbrecht: Maybe. I asked in an issue at Lux.jl if they want to include it there, and I didn't get a response yet. Without doc strings / documentation and tests it's not even 100 lines of code. 

On top of that, I would also at least do one more example with it, so that we are really sure it's what we need/want.

test/reactant/Project.toml
maximilian-gelbrecht: Yeah, it should work. We do only test with 1.12 though, also for the sake of reducing CI time. 

Also in general I think the only reason that I see to hang onto 1.10 is vanilla Enzyme still working better.

ext/TerrariumReactantExt/transfer.jl
maximilian-gelbrecht: Ok, I was already able to remove this now by defining a `default_clock` function that get's extended here instead.

src/processes/surface_hydrology/canopy_interception/canopy_interception.jl
maximilian-gelbrecht: ok, I've included the `safediv` now, seems reasonable.

bgroenks96:
I will checkout your branch locally and fix the conflicts.

maximilian-gelbrecht:
I think I've already done that now. I hope it works at least

maximilian-gelbrecht:
Ok, all tests passing again.

bgroenks96 (approved):

Issue #141, opened 2026-07-11
Simple single-layer snow scheme with SEB coupling


bgroenks96:
Both seasonal and perennial snow cover is a key element of the land surface, especially in high-elevation and polar regions. It plays a key role in both the water cycle (snow melt) and energy balance (albedo, storage, and insulation) of the land surface.  As such, it constitutes a major gap preventing Terrarium from being useful for global simulations.

We plan to start with a simple, single-layer snow scheme loosely based on the Utah Energy Balance (UEB) described by [Tarboton, Chowdhury & Jackson (1994)](https://digitalcommons.usu.edu/water_rep/60/). This models the snowpack as a single set of vertically integrated prognostic state variables for the mass and energy balance.

The following text is an excerpt from the implementation plan generated by Claude 🤖 :

## Background: the Utah Energy Balance (UEB) model

Tarboton, Chowdhury & Jackson (1994) describe a lumped, depth-averaged, single-layer snowpack with
two prognostic state variables: water equivalence `W` (m) and energy content `U` (kJ/m²) relative to
ice at 0 °C. The governing ODEs (their eqns 1–2) are

```
dU/dt = Q_sn + Q_li + Q_p + Q_g − Q_le + Q_h + Q_e − Q_m
dW/dt = P_r + P_s − M_r − E
```

with net shortwave `Q_sn`, incoming/outgoing longwave `Q_li`/`Q_le`, advected precipitation heat
`Q_p`, ground heat flux `Q_g`, sensible/latent fluxes `Q_h`/`Q_e`, advected meltwater heat `Q_m`,
rain/snow inputs `P_r`/`P_s`, meltwater outflow `M_r`, and sublimation `E`. Depth-averaged snow
temperature follows from `(U, W)` by phase partitioning (eqns 6–8); the **snow surface temperature**
`T_s` is solved separately from an equilibrium surface energy balance (eqns 20–22) because snow is a
strong insulator and `T_s ≠ T_avg`. Meltwater outflow uses a Darcy relation on the liquid fraction
in excess of capillary retention (eqns 23–24).

### Adaptation to Terrarium

- UEB's predictor–corrector time integration is **not** carried over. Terrarium supplies the
  integrator; the snow scheme contributes only well-formed ODE tendencies via `compute_tendencies!`.
- UEB's effective soil interaction depth `D_e` (which lumps a 0.4 m soil slab into `U`) is **dropped**.
[...]

PR #142, opened 2026-07-11
Separate ground and soil heat fluxes

bgroenks96:
This PR is a prerequisite for #141.

The basic problem is that currently `ground_heat_flux` is used directly as the upper boundary condition for the soil. However, in the case that snow is present, this would not be correct, since the ground heat flux in the SEB would instead be applied to the snow surface.

This PR separates the ground heat flux from the "soil heat flux" so that these can be later distinguished. Since snow is not implemented yet, the current implementation still just sets `soil_heat_flux = ground_heat_flux`.

## Additional changes
- Adds new workflow for implementation plans to `AGENTS.md`
- Adds current and previous plan documents to `docs/dev`


docs/src/models/soil_model.md
```diff
@@ -108,7 +108,7 @@ Terrarium provides a set of named boundary condition aliases for the most common
 ### Energy boundary conditions
 
 ```@docs; canonical = false
-GroundHeatFlux
+SoilHeatFlux
```
maximilian-gelbrecht, line 111: Stupid questions but language-wise in land modelling what would you say is the difference between _ground_ and _soil_

src/models/coupled/land_model.jl
```diff
@@ -58,11 +58,14 @@ function StateVariables(
     # Initialize BC fields for coupling
     ground_heat_flux = initialize(vars.auxiliary.ground_heat_flux, grid, clock, boundary_conditions, fields)
     infiltration = initialize(vars.auxiliary.infiltration, grid, clock, boundary_conditions, fields)
-    ground_heat_flux_bc = GroundHeatFlux(ground_heat_flux)
+    # Without snow, the heat flux into the soil top equals the surface energy balance closure flux
+    # (`ground_heat_flux`); wire the soil-top BC directly to it. When a snow component is added,
+    # the soil BC will instead be wired to a distinct `soil_heat_flux`.
```
maximilian-gelbrecht, line 63: is what the comment says automatically done by the constructor or does the user has to do anything?

maximilian-gelbrecht (commented):
I am a bit confused. So this PR is also just renaming? 

In general I appreciate separating the implementation into smaller PRs though! So thank you for doing that here as well.

AGENTS.md
maximilian-gelbrecht, line 1: I am fine with that! Just add the folder structure as well if you want it ordered by `YYYY-MM` folders as you've done here

docs/src/models/soil_model.md
bgroenks96: It depends who you ask. It's a question which constantly bothers me as well. From the perspective of the atmosphere, everything below the "surface" (end of the atmosphere) could be considered "ground", even the canopy. But from the perspective of the land, snow is not really "ground" because, except in a few places, it's largely ephemeral and has a fundamentally different material structure than soil/rock.

In Terrarium, I guess we are implicitly adopting the former view that snow is still effectively part of the "ground" in the more abstract sense because this just makes the terminology surrounding the surface energy balance easier to keep track of. This PR basically disambiguates "ground heat flux", meaning the exchange between the solid (non-vegetated) land surface and the atmosphere, from "soil heat flux", which is the heat flux into the *soil* (or I guess rock, as a special case).

It's all very confusing. But maybe @olivierbonte also has opinions on this.

src/models/coupled/land_model.jl
bgroenks96: The constructor does this. The user of `LandModel` doesn't need to do anything.

maximilian-gelbrecht: But the constructor doesn't do it yet because the snow model isn't there yet, or not?

docs/src/models/soil_model.md
maximilian-gelbrecht: Ok, thanks makes sense.

src/models/coupled/land_model.jl
bgroenks96: Yes. See #144.

maximilian-gelbrecht (approved):

AGENTS.md
bgroenks96: Fixed in [ee80a88](https://github.com/NumericalEarth/Terrarium.jl/pull/142/commits/ee80a88340db53bf32c25348df9e107207f41963)

PR #150, opened 2026-07-22
Add prescribed and prognostic phenology schemes

### LINKED ISSUE #146: Prescribed phenology scheme
The easiest way to get a vegetation model (if you can call it that)  working with our current implementation is to add a `PrescribedPhenology` scheme that allows the user to set `LAI` as a constant or time-varying `input` variable.

The implementation should be fairly straightforward:
- `PrescribedPhenology` provides `LAI` as an input variable
- The `phenology_factor` is dropped/ignored
- Other vegetation carbon processes such as photosynthesis, as well as hydrology processes like canopy interception, can consume the phenology process and/or the `LAI` variable
- Carbon and vegetation dynamics must be set to `nothing` in `VegetationCarbon`

In addition, we will need input data for LAI. I have already downloaded ERA5-Land LAI from 1980-2024 and can regrid this to either 10km or 1 deg resolution depending on where we want to store it (see also #80).

### LINKED ISSUE #149: Implement prognostic phenology scheme adapted from PALADYN
The current implementation of `PALADYNPhenology` is just a stub. A key question for implementing the original scheme from PALADYN is how to handle the dependence of the phenology factor on seasonal growing degree days (GDD) which are accumulated over the course of the growing season. The proposal here is to get around this by treating GDD as a prognostic variable that is integrated over time with an added decay term when the air temperature drops below a certain threshold:

```math
\frac{d \text{GDD}}{d t} = \max(T_{\text{air}} − T_\text{base}, 0) − (\text{GDD} / \tau) \cdot \mathcal{1}[T_{\text{air}} < T_{\text{GDD}}]
```

The following is a summary of the problem generated by Claude :robot:.

## Problem description

The autotrophic respiration scheme `PALADYNAutotrophicRespiration` requires a `phenology_factor`
input `ϕ ∈ [0, 1]` (leaf-out fraction), consumed in the root-maintenance-respiration term
(`src/processes/vegetation/respiration/autotrophic_respiration.jl`, `compute_Rm`). This factor must
be produced by whichever `AbstractPhenology` implementation is coupled into the vegetation model.

Two phenology implementations exist:

- `PrescribedPhenology` — treats `leaf_area_index` as a (possibly time-varying) input. Prior to this
  work it produced *no* `phenology_factor`, so pairing it with `PALADYNAutotrophicRespiration` left
  that input unsatisfied.
- `PALADYNPhenology` — currently a stub. `compute_phenology_factor` returns a hardcoded `1.0` and
  `compute_f_deciduous` returns `0.0` (evergreen). The real PALADYN scheme is GDD-based and thus
  history-dependent; a naive implementation would require storing previous timesteps' temperature
  inputs, which conflicts with the differentiable, continuous-time design of Terrarium.

This plan specifies how each scheme derives `ϕ` in a manner consistent with the project rules:
continuous-time dynamics only, no discrete/per-day updates, no stored input history, GPU- and
Enzyme-compatible.

### PALADYN phenology (Willeit & Ganopolski, 2016; after Sitch et al., 2003)

Phenology type is set by the coldest-month temperature: if it falls below a PFT-specific threshold
`T_phen_cmon`, the PFT is cold-deciduous, otherwise evergreen. Needleleaf trees are always evergreen.
For deciduous PFTs:

- `LAI = ϕ · LAI_b`  (balanced/maximum LAI `LAI_b`).                                        (eq. 82)
- `ϕ = gdd / gddcrit`, where `gdd` accumulates degree-days above a base temperature `T_gdd_base`
  at a PFT-specific critical value `gddcrit`.                                                (eq. 83)
- After `ϕ` reaches its maximum of 1 it is held constant until air temperature drops below
  `T_gdd_base`. Senescence then proceeds, with all leaves lost at `T_gdd_base − 5 °C`.
- Raingreen (moisture-driven) phenology is not represented.

The scheme has three history-dependent elements: (1) the GDD accumulation, (2) the implicit annual
GDD reset used by LPJ/Sitch, and (3) the coldest-month-temperature classification. The senescence
ramp is an instantaneous function of current air temperature and needs no history.

### Design constraint

Terrarium prohibits discrete-time updates (e.g. "reset once per year", "sum once per day"). All
memory of the past must live in **prognostic state variables** integrated by the timestepper, and
all dynamics must be well-posed ODEs. This is what lets us discard stored input history.

### LINKED ISSUE #159: A simple vegetation model configuration based on prescribed LAI
We want to have a simple vegetation model configuration based on prescribed LAI, where only photosynthesis,
stomatal conductance and autotrophic respiration are active.  In this configuration we would have:

- LAI is prescribed via `PrescribedPhenology`.
- No vegetation carbon dynamics: no prognostic carbon pools are computed.
- No vegetation dynamics: no prognostic vegetation area fractions are computed.

For this to work we need to make one assumption: `LAI_b = LAI_max`, which means setting
the balanced leaf area index equal to the PFT-specific maximum LAI parameter, rather than computing it
from the prognostic carbon pools.

This requires an additional refactoring of autotrophic respiration:

- Maintenance respiration should be formulated in terms of `LAI_b` rather than `C_veg`,
so that the formulation is valid in both cases: when `LAI_b` is derived from the carbon pools,
and in the simplified case where `LAI_b = LAI_max`. For `b_wl = 1`, the relationship between the vegetation carbon pool and the balanced leaf area can be formulated as:

$$C_{\text{veg}} = \left(\frac{2}{\text{SLA}} + a_{wl}\right)\text{LAI}_b$$

- `SLA`, `a_wl` and `LAI_max` need to be relocated so that they can be used by both respiration and carbon dynamics.

bgroenks96:
This PR adds both a `PrescribedPhenology` scheme and implements the modified phenology scheme from PALADYN as described in #149.

### Open questions
- The `PrescribedPhenology` scheme currently just declares LAI as an input variable, which means its up to the user to define a `FieldTimeSeries` or other `InputSource` that is time-varying. Should we provide a pre-specified `InputSource` for this? Or possibly a separate phenology type `SeasonalPhenology` that hard codes a seasonal cycle based on some parameters?
- How should `PrescribedPhenology` interact with PFTs? (out of scope for this PR but still important question to answer)

### Additional changes
- Adds `PrescribedVegetation` as alternative coupled vegetation process that omits autotrophic respiration, carbon dynamics, and vegetation dynamics
- Adds `PlantTraits` which refactors a subset of PFT-specific parameters into a separate type. This is a precursor to full PFT support.
- Adds a new ERA5-Land LAI asset
- Adds a new global vegetation example script
- Refactors the `Rasters` extension and adds support for cycling time series (climatologies)
- Improved API for defining derived auxiliary variables as `KernelFunctionOperation`s

Resolves #146
Resolves #149
Resolves #159


src/processes/vegetation/phenology/paladyn_phenology.jl
```diff
@@ -0,0 +1,188 @@
...
+
+Compute the phenology factor `ϕ` [-] from accumulated growing degree days `gdd` and air temperature `T_air`.
+
+The cold-deciduous factor combines the growth-phase ramp `gdd / gdd_crit`
+[willeitPALADYNV10Comprehensive2016; Eq. (83)](@cite) with a senescence ramp that declines linearly from 1
+at `T_gdd_base` to 0 at `T_gdd_base − T_senescence_range`. Taking the minimum reproduces all three PALADYN
+regimes (linear green-up, held at 1 when mature and warm, temperature-driven senescence). The result is then
+blended with the evergreen value of 1 via the prescribed deciduous fraction.
+"""
+@inline function compute_phenology_factor(phenol::PALADYNPhenology{NF}, gdd, T_air) where {NF}
```
bgroenks96, line 73: @matteo-willeit Please review

src/processes/vegetation/phenology/paladyn_phenology.jl
```diff
@@ -0,0 +1,188 @@
...
+"""
+    $SIGNATURES
+
+Compute the growing-degree-day tendency `d(gdd)/dt` [K⋅day/s] at a single grid point.
+
+Heat above the base temperature is accumulated (in degree-days per second) and, below the base temperature,
+the accumulator relaxes toward zero with timescale `gdd_relaxation_time`. The relaxation provides a
+continuous surrogate for the discrete annual reset of the original scheme.
+"""
+@inline function compute_gdd_tendency(phenol::PALADYNPhenology{NF}, gdd, T_air) where {NF}
```
bgroenks96, line 104: @matteo-willeit Please review

src/processes/vegetation/phenology/paladyn_phenology.jl
```diff
@@ -0,0 +1,188 @@
...
+[willeitPALADYNV10Comprehensive2016](@cite) and [sitchEvaluationEcosystemDynamics2003](@cite).
+
+The instantaneous leaf area index is `LAI = ϕ·LAI_b`, where `LAI_b` is the balanced (annual-maximum)
+leaf area index and `ϕ` is the phenology factor. For a cold-deciduous PFT, `ϕ` ramps up linearly with
+accumulated growing degree days above a base temperature `T_gdd_base` at a rate set by `gdd_crit`
+[willeitPALADYNV10Comprehensive2016; Eq. (83)](@cite), holds at 1 once fully leafed out, and declines
+linearly during senescence as air temperature falls from `T_gdd_base` to `T_gdd_base − T_senescence_range`.
+
+Unlike the discrete daily accumulation of the original scheme in PALADYN, the growing degree days are integrated as
+a prognostic state variable, so no history of past inputs need be stored. To recover a periodic seasonal
```
maximilian-gelbrecht:
```suggestion
a prognostic state variable, so no history of past inputs needs to be stored. To recover a periodic seasonal
```

src/processes/vegetation/phenology/prescribed_phenology.jl
```diff
@@ -0,0 +1,30 @@
+"""
+    $TYPEDEF
+
+Prescribed vegetation phenology where `leaf_area_index` is treated a (possibly time-varying) input variable.
```
maximilian-gelbrecht:
```suggestion
Prescribed vegetation phenology where `leaf_area_index` is treated as a (possibly time-varying) input variable.
```

maximilian-gelbrecht:
```suggestion
Prescribed vegetation phenology where `leaf_area_index` is treated as a (possibly time-varying) input variable.
```

src/processes/vegetation/stomatal_conductance/medlyn_stomatal_conductance.jl
```diff
@@ -47,12 +47,6 @@ Includes minimum conductance and light extinction effects based on LAI, scaled b
         photo::LUEPhotosynthesis{NF},
         vpd, An, co2, LAI, β
     ) where {NF}
-    # Preconditions
-    @assert isfinite(vpd) && abs(vpd) > zero(NF) "vapor pressure deficit must be greater than zero"
```
maximilian-gelbrecht, line 51: Did you remove this with Reactant in mind or for another reason?

src/processes/vegetation/vegetation_carbon_cycle.jl
```diff
@@ -81,21 +81,21 @@ function compute_auxiliary!(
     # PAW: needs soil saturation profile and computes soil_moisture_limiting_factor
     compute_auxiliary!(state, grid, veg.plant_available_water, soil)
 
-    # Veg. carbon dynamics: needs C_veg(t-1) and computes LAI_b(t-1)
+    # Veg. carbon dynamics: needs C_veg(t) and computes LAI_b(t)
```
maximilian-gelbrecht, line 97: Is this also a change in the parametereziation? Were we running it on the previous timestep before? If so, how?

src/utils/kernel_utils.jl
```diff
@@ -1,3 +1,42 @@
+struct KernelFunction{autonomous, Func, Args}
```
maximilian-gelbrecht, line 1: This looks good to me on first sight, but I still find it hard to judge if things like this also always works easily with GPU and Reactant (so with fully static dispatch). Did you check?

maximilian-gelbrecht (commented):
I can't really comment much on the physics directly, so it's just a few typos and questions. 

> The PrescribedPhenology scheme currently just declares LAI as an input variable, which means its up to the user to define a FieldTimeSeries or other InputSource that is time-varying. Should we provide a pre-specified InputSource for this? Or possibly a separate phenology type SeasonalPhenology that hard codes a seasonal cycle based on some parameters?

I think some convenience constructors would be really good. 
Would you put the data here, in another repo or in NumericalEarth? 

With regards for a `SeasonalPhenology`, do you see much of a purpose for that when we have a full prognostic vegetation model coupled or would that just be a stop-gap solution

src/processes/vegetation/stomatal_conductance/medlyn_stomatal_conductance.jl
maximilian-gelbrecht: Speaking of Reactant, `FieldTimeSeries` will be needed here, so this will be something we need to tackle separately probably as this one is a bit more tricky, I think.

src/processes/vegetation/vegetation_carbon_cycle.jl
bgroenks96: No, the comment was just wrong, I think.

docs/dev/2026-07/2026-07-22_PLAN_01_prognostic_phenology.md
```diff
@@ -0,0 +1,253 @@
...
+This plan specifies how each scheme derives `ϕ` in a manner consistent with the project rules:
+continuous-time dynamics only, no discrete/per-day updates, no stored input history, GPU- and
+Enzyme-compatible.
+
+## Background
+
+### PALADYN phenology (Willeit & Ganopolski, 2016; after Sitch et al., 2003)
+
+Phenology type is set by the coldest-month temperature: if it falls below a PFT-specific threshold
+`T_phen_cmon`, the PFT is cold-deciduous, otherwise evergreen. Needleleaf trees are always evergreen.
```
matteo-willeit, line 111: Needleleaf trees can also be deciduous.

docs/dev/2026-07/2026-07-22_PLAN_01_prognostic_phenology.md
```diff
@@ -0,0 +1,253 @@
...
+>   autonomous variant that does not pass the clock to the kernel function.
+> - `src/processes/vegetation/phenology/prescribed_phenology.jl`: `variables(phenol::...)` binds the
+>   process instance and passes it through `kernel(phenology_factor, phenol)` so the kernel function
+>   receives the parameters. Verified: for prescribed `LAI = 2.0`, `max_leaf_area_index = 4.0`, the
+>   computed `phenology_factor` is `0.5`. Full vegetation, state-variable, and abstract-variable test
+>   suites pass (the one `parameter_handling` failure is pre-existing and unrelated).
+>
+> 2026-07-22 — Part B implemented. `PALADYNPhenology` rewritten as a prognostic GDD scheme:
+> - New parameters `T_gdd_base`, `gdd_crit`, `T_senescence_range`, `gdd_relaxation_time`,
+>   `f_deciduous` (placeholder defaults; PALADYN Table 5 values flagged TODO).
```
matteo-willeit, line 39: Not sure I understand what `f_deciduous `actually represents. A PFT is either deciduous or not, no?

docs/dev/2026-07/2026-07-22_PLAN_01_prognostic_phenology.md
```diff
@@ -0,0 +1,253 @@
...
+
+- Growth (`T_air ≥ T_gdd_base`): temperature term saturates at 1 ⇒ `ϕ = gdd/gddcrit` (eq. 83).
+- Mature (`gdd/gddcrit ≥ 1`, warm): `ϕ = 1`, held.
+- Senescence (`T_air < T_gdd_base`): temperature term falls below 1 and draws `ϕ` linearly to 0 at
+  `T_gdd_base − 5 °C`.
+
+**Deciduous/evergreen classification:** the coldest-month-temperature test is an annual statistic
+that does not reduce to a clean ODE. Represent `f_deciduous` as a prescribed PFT/site property (or an
+infrequently-updated input), consistent with the current `compute_f_deciduous` stub and with the fact
+that needleleaf PFTs are always evergreen. Combine as in the existing `compute_LAI`:
```
matteo-willeit: Again, needleleaf PFTs can be deciduous as well (e.g. Larix)

docs/dev/2026-07/2026-07-22_PLAN_01_prognostic_phenology.md
```diff
@@ -0,0 +1,253 @@
...
+These are pre-existing TODOs in the vegetation code, surfaced here for tracking; they are **out of
+scope** for this plan unless noted:
+
+- `paladyn_phenology.jl` — replace the stubbed `compute_f_deciduous` (returns 0) and
+  `compute_phenology_factor` (returns 1) with the implementation above; add the phenology parameters.
+- `respiration/autotrophic_respiration.jl`:
+  - `cn_sapwood`, `cn_root` — physical meaning and units unconfirmed (`# TODO check ... + add unit`).
+  - `compute_f_temp` hardcodes the constants `308.56`, `56.02`, `46.02`; per the rule against
+    hardcoded literals these should move into the process struct or `PhysicalConstants`.
+  - The `T_soil > 7 °C` hard bound is inherited from CLIMBER-X/PALADYN and is not further justified;
```
matteo-willeit: Where do these 7°C come from?

src/processes/vegetation/phenology/paladyn_phenology.jl
matteo-willeit: Looks good

matteo-willeit (commented):
Looks generally good to me, I just added a few comments.

docs/dev/2026-07/2026-07-22_PLAN_01_prognostic_phenology.md
bgroenks96: Yes, true.

bgroenks96: I think the point of it was to represent the fraction of the grid cell covered by deciduous vegetation since we don't currently explicitly handle PFTs. I suppose this would be removed or replaced with a state variable once we have PFTs.

docs/dev/2026-07/2026-07-22_PLAN_01_prognostic_phenology.md
bgroenks96: I think it's like this in the CLIMBER-X code, no?

https://github.com/cxesmc/climber-x/blob/60b1818b9eb57aaf5a318f071a146f02479db09a/src/lnd/photosynthesis.f90#L142

bgroenks96: I think we will need to replace this with a learned smooth function at some point but fine to keep the threshold for now.

matteo-willeit: I see, but the threshold is 240 K (or -33°C). That threshold is anyway there only to save on the computation of the `exp `function, because f_temp will be very close to zero at those low temperatures anyway. So I think that you can safely remove the threshold.

bgroenks96: Hmm, sorry, I think I misread the code there. I thought I remembered there being a threshold somewhere with 280 K, but maybe I'm wrong. @mahabadri I think you originally wrote the autotrophic respiration code; was there a threshold or not?

src/utils/kernel_utils.jl
bgroenks96: It's not used within kernels.

docs/dev/2026-07/2026-07-22_PLAN_01_prognostic_phenology.md
bgroenks96: See the autotrophic respiration code here which references PALADYN (I guess this where Claude got this from) but doesn't provide a direct reference:

https://github.com/NumericalEarth/Terrarium.jl/blob/main/src/processes/vegetation/respiration/autotrophic_respiration.jl#L58

bgroenks96: Nevermind, I checked the git history, and I was the one who added it. I guess I also confused the temperature value then too. Sorry for the mistake, will remove the threshold.

bgroenks96: Fixed in [f0a430a](https://github.com/NumericalEarth/Terrarium.jl/pull/150/commits/f0a430a7b4585ff9c4033f783c5b5238b0e158a6)

bgroenks96:
@maximilian-gelbrecht The current approach to the `InputSource` is to use a `RasterInputSource` with `cycle=true` (added in this PR). No custom `SeasonalPhenology` or `InputSource` needed (yet).

test/inputs/raster_inputs.jl
```diff
@@ -18,18 +21,19 @@ const RasterInputSource = TerrariumRastersExt.RasterInputSource
...
+    nx, ny = length(lon), length(lat)
+    static_data = rand(Float32, nx, ny)
+
     @testset "Static raster input" begin
         # Create a dummy NetCDF file with static (no time dimension) data
         static_file = joinpath(testdir, "static_data.nc")
 
-        # Generate random data matching the ring grid size
-        # Use 2D layout with X and Y dimensions (flattened to match ring grid)
-        nx, ny = 32, 24  # arbitrary 2D shape that gives npoints when flattened
```
bgroenks96, line 30: @maximilian-gelbrecht This is the mysterious incorrect Raster test. `nx * ny` does not match the number of points in the `FullHEALPixGrid`... so I don't really understand how this ever worked.

ext/TerrariumRastersExt/TerrariumRastersExt.jl
```diff
@@ -34,126 +48,181 @@ struct RasterInputSource{NF, name, VD, TT, IM <: AbstractVector{Int}, RS <: Abst
...
+    t1 = timestamp(source.reftime, first(timedim.val))
+    tN = timestamp(source.reftime, last(timedim.val))
+    ti = timestamp(source.reftime, clock.time)
+    Δt₀ = (tN - t1) / (N - 1)
+    period = tN - t1 + Δt₀ # approximate for non-equidistant spacing
+    # search for indices between which t lies, converting everything to relative time in seconds;
+    # note that we use clock.time again here because searchsorted applies by to the second argument
+    indexes = searchsorted(timedim.val, clock.time, by = t -> t1 + mod(timestamp(source.reftime, t) - t1, period))
+    lower, upper = last(indexes), first(indexes)
+    @inbounds if t1 < ti < tN || source.extrapolation isa Cyclical
```
maximilian-gelbrecht, line 160: It's rather an optimisation for another time but for cyclical, so yearly climatology files, it would probably also work well to just interpolate to the time step once in the beginning (assuming a fixed time step)

src/input_output/assets.jl
```diff
@@ -6,13 +7,20 @@ File extension (including the leading dot) used by files of the given [`FileForm
...
     $TYPEDEF
 
 Lightweight base type for Terrarium land data assets.
 """
 abstract type AbstractLandAsset end
 
+"""
+    $TYPEDEF
+
+Time-invariant spatial fields from ERA5-Land at the native 0.1° x 0.1° resolution.
```
maximilian-gelbrecht, line 22: Could you be a bit more explicit what invariants?

src/input_output/assets.jl
```diff
@@ -29,6 +55,8 @@ format(::ERA5LandForcings) = NetCDF()
 indices(::ERA5LandForcings) = (:, :, :)
 native_grid(::ERA5LandForcings) = RingGrids.FullGaussianGrid(72)
 
+# get_asset
```
maximilian-gelbrecht, line 88: Also here which meteorological variables

src/processes/vegetation/stomatal_conductance/medlyn_stomatal_conductance.jl
```diff
@@ -27,7 +27,9 @@ MedlynStomatalConductance(::Type{NF}; kwargs...) where {NF} = MedlynStomatalCond
 
 variables(::MedlynStomatalConductance) = (
     auxiliary(:canopy_water_conductance, XY(), units = u"m/s"), # Canopy conducatance for water vapor
-    auxiliary(:leaf_to_air_co2_ratio, XY()), # Ratio of leaf-internal and air CO2 concentration [-]
+    input(:leaf_area_index, XY(), units = u"m^2/m^2"),
```
maximilian-gelbrecht, line 30: is the `MedlynStomatalConductance` then just defined for the prescribed case?

src/processes/vegetation/plant_traits.jl
```diff
@@ -0,0 +1,39 @@
+"""
+    $TYPEDEF
+
+Plant-specific parameters for a single functional type. This is a temporary, partial construct
+that will be soon be replaced will full support for PFTs and other trait parameterizations.
```
maximilian-gelbrecht, line 5: As I am a bit oblivious to vegetation models: what do PFTs do beyond that here (aside from multiple PFTs existing)?

src/processes/vegetation/stomatal_conductance/medlyn_stomatal_conductance.jl
maximilian-gelbrecht: Or it's rather here because having it as an input is the minimum and having it as a diagnostic or prognostic would be defined by another process which then overrules this definition here, right?

test/inputs/raster_inputs.jl
maximilian-gelbrecht: huh, yeah that's strange. I also don't understand this

maximilian-gelbrecht (commented):
looks good, just a few very minor comments

ext/TerrariumRastersExt/TerrariumRastersExt.jl
bgroenks96: Yes. I guess you can also already just do this offline manually to avoid this interpolation step.

src/processes/vegetation/stomatal_conductance/medlyn_stomatal_conductance.jl
bgroenks96: Yes to the latter.

src/processes/vegetation/plant_traits.jl
bgroenks96: This struct still assumes a single PFT. The future implementation will generalize that.

bgroenks96: But hopefully the interface will also be flexible enough to support trait-based modeling without explicit PFTs.

src/input_output/assets.jl
bgroenks96: Done in [aca06cb](https://github.com/NumericalEarth/Terrarium.jl/pull/150/commits/aca06cbaec925be89c2d9c3fe13174d0221c44cf)

src/input_output/assets.jl
bgroenks96: Done in [aca06cb](https://github.com/NumericalEarth/Terrarium.jl/pull/150/commits/aca06cbaec925be89c2d9c3fe13174d0221c44cf)

src/processes/vegetation/dynamics/carbon_dynamics.jl
```diff
@@ -61,15 +46,18 @@ vegetated area and spreading of the given PFT based on the balanced Leaf Area In
...
-    elseif LAI_b <= vegcarbon_dynamics.LAI_max
-        λ_NPP = (LAI_b - vegcarbon_dynamics.LAI_min) /
-            (vegcarbon_dynamics.LAI_max - vegcarbon_dynamics.LAI_min)
-    else
-        λ_NPP = NF(1.0)
-    end
+@inline function compute_λ_NPP(vegcarbon_dynamics::PALADYNCarbonDynamics{NF}, traits::PlantTraits{NF}, LAI_b) where {NF}
+    LAI_min = traits.minimum_leaf_area_index
+    LAI_max = traits.maximum_leaf_area_index
+    λ_NPP = ifelse(
```
bgroenks96, line 52: @maximilian-gelbrecht I guess this should make this function now Reactant-compatible, yes?

test/inputs/raster_inputs.jl
bgroenks96: 🤷‍♂️

src/processes/vegetation/dynamics/carbon_dynamics.jl
maximilian-gelbrecht: In principle yes. 

I'll look at the Reactant compat of this some time after JuliaCon I think. The input data interpolation might be a problem that needs a bit of thought.

maximilian-gelbrecht (approved):

PR #152, opened 2026-07-28
Add native support for automatically downloading input files

### LINKED ISSUE #80: Input Data Handling
How do we want to handle input data? 

At the moment, there's e.g. the data missing to reproduce the global soil script just from the repo. For the Pluto notebook in #77 we also need more input data. 

Short term we could just put all in `inputs` folder. Do we want that for now? 

Long term we could use the artefact system in combination with a separate repository just for the data, as done now for Speedy in https://github.com/SpeedyWeather/SpeedyWeather.jl/pull/983
When we do a new repository for input data, we could/should probably also wait if we still move to NumericalEarth org.

bgroenks96:
This PR adds a `Artifacts.toml` file for registering and lazily downloading large input data files, along with a minimal API for handling them.

I would regard this as still a somewhat temporary solution that may become obsolete as more datasets are added to NumericalEarth.jl. The future Terrarium extension for NumericalEarth could also add hooks so that the same Terrarium `get_asset`/`load_asset` interface can be used with NumericalEarth `Metadataset`s.

Artifacts are currently added manually via [ArtifactUtils.jl](https://github.com/JuliaPackaging/ArtifactUtils.jl) which should be installed in the developer's root environment:

```julia
using ArtifactUtils

add_artifact!("Artifacts.toml", name, tarball_url)
```

## Summary

Replaces checked-in NetCDF input files with a lazy, artifact-based data-access layer. Reference datasets are described by lightweight asset types and downloaded on demand from a static Artifacts.toml, keeping large binaries out of the repository.

## Changes

Asset types (`src/input_output/get_asset.jl`): add `AbstractLandAsset` with `ERA5LandInvariants` and `ERA5LandForcings`. Each carries metadata — `artifact_name`, `varnames`, `format`, `indices`, and `native_grid`.
Artifact retrieval: get_artifact resolves an entry in the package Artifacts.toml and installs it via `Pkg.Artifacts.ensure_artifact_installed` (downloads and unpacks the .tar.gz). `get_asset(asset)` returns the path to the extracted file, deferring raw reading to the caller.
Convenience loader: `load_asset(asset, name; NF, fill_value)` returns a `RingGrids.Field` on the asset's `native_grid`. The NetCDF reading and field wrapping are implemented in `TerrariumRastersExt` (via `Rasters`), including `reconcile_latitudes`, which drops the two pole rings so a pole-inclusive lon–lat grid maps onto a `RingGrids` full grid, and indices-based slice selection.
Artifacts.toml: static, lazy entries for era5-land-invariants and era5-land-forcings-N72 (populated offline with `ArtifactUtils.jl`).
Cleanup: remove inputs/era5-land_land_sea_mask_N{72,145}.nc; add Downloads and Pkg to [deps]/[compat].
Example: `examples/simulations/soil_heat_global_era5.jl` updated to load the land–sea mask and forcings through the new API.
Incidental: Reactant extension moved into a subfolder.

## Notes
load_asset handles static (2-D) fields; indices selects a single time record for the invariants ((:, :, 5)) and retains all records for the forcings.
No NCDatasets extension is added — NetCDF I/O reuses the existing Rasters path.
🤖 Generated with [Claude Code](https://claude.com/claude-code)

Resolves #80


maximilian-gelbrecht (approved):
Looks good to me, no direct comments to the code 

I think it's a good start for now, and then we can still decide if we extend this here (e.g. with automatic interpolation or other such options) or just rely more heavily on NumericalEarth

PR #59, opened 2025-12-16
Global sensitivity example

maximilian-gelbrecht:

PR #153, opened 2026-07-29
Remove NF from field types (Reactant compat) 

maximilian-gelbrecht:
This came up in #59 and is needed to deal with the fact that Reactant might replace individual `NF` with traced versions of it. 

Should we also directly address the `AbstractParameterzation` and `@adapt_structure` here, @bgroenks96 ?

### DIFF
--- a/src/processes/soil/stratigraphy/soil_horizon.jl
+++ b/src/processes/soil/stratigraphy/soil_horizon.jl
@@ -4,18 +4,23 @@
 Represents an arbitrary soil horizon whose properties (texture and porosity) are assumed to
 be constant across both space and time.
 """
-@parameterized struct ConstantSoilHorizon{NF, name, Porosity <: AbstractSoilPorosity{NF}} <: AbstractSoilHorizon{NF, name}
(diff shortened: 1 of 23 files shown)


maximilian-gelbrecht:
Actually this isn't the right approach yet. I have to look into this again.

src/processes/soil/stratigraphy/soil_horizon.jl
```diff
@@ -38,14 +43,14 @@ end
 Represents an arbitrary soil horizon whose properties (texture and porosity) are prescribed
 via input `Field`s and can therefore vary across space and (less commonly) time.
 """
-@parameterized struct PrescribedSoilHorizon{NF, name, Porosity <: AbstractSoilPorosity{NF}} <: AbstractSoilHorizon{NF, name}
```
bgroenks96, line 41: Yeah do not like... why is this necessary?

bgroenks96:
Was this superseded by #157?

maximilian-gelbrecht:
No, that's independent. I still close this because handling the merge conflicts is more complicated than starting fresh in this case

PR #155, opened 2026-07-31
Add missing dispatches for prescribed SEB

bgroenks96:
Needed for integration with NumericalEarth; see NumericalEarth/NumericalEarth.jl#148

### DIFF
--- a/src/processes/surface/surface_energy_balance.jl
+++ b/src/processes/surface/surface_energy_balance.jl
@@ -71,7 +71,7 @@ and surface hydrology state.
 """
 function solve_surface_energy_balance!(
         state, grid,
-        seb::SurfaceEnergyBalance{NF, <:ImplicitSkinTemperature},
+        seb::SurfaceEnergyBalance{NF},
         constants::PhysicalConstants,
         atmos::AbstractAtmosphere,
         hydrology::Optional{AbstractSurfaceHydrology} = nothing,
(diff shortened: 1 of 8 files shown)


src/processes/surface/surface_energy_balance.jl
```diff
@@ -87,6 +87,29 @@ function solve_surface_energy_balance!(
     return nothing
 end
 
+"""
+    $TYPEDSIGNATURES
+
+Close the surface energy balance when the skin temperature is prescribed (e.g. supplied by an external
+coupler such as an atmosphere-land flux scheme).
+"""
+function solve_surface_energy_balance!(
```
maximilian-gelbrecht: seems like a very un-Julia name, with `SurfaceEnergyBalance` already in the list of arguments, couldn't it just be `solve!`?  Or would that be to ambiguous?

src/processes/surface/surface_energy_balance.jl
```diff
@@ -87,6 +87,29 @@ function solve_surface_energy_balance!(
...
+        constants::PhysicalConstants,
+        atmos::AbstractAtmosphere,
+        args...
+    ) where {NF}
+    # Diagnose the radiative budget (net radiation) from the prescribed skin temperature and albedo.
+    compute_auxiliary!(state, grid, seb.radiative_fluxes, seb, constants, atmos)
+    # Diagnose the turbulent fluxes, if necessary
+    compute_auxiliary!(state, grid, seb.turbulent_fluxes, seb, constants, atmos)
+    # Close the energy balance by computing ground heat flux from the radiative and turbulent fluxes
+    compute_ground_heat_flux!(state, grid, seb.skin_temperature, seb)
```
maximilian-gelbrecht: I try to understand the difference between the prescribed and un-prescribed version. I am a bit surprised the difference in code is so large. I can see the ground heat flux computation below in the `compute_surface_energy_fluxes!`, but wants difference then between how the fluxes are diagnosed here, and how they are computed by `compute_surface_energy_fluxes!`?

src/processes/surface/turbulent_fluxes.jl
```diff
@@ -15,6 +15,9 @@ variables(::PrescribedTurbulentFluxes) = (
     input(:latent_heat_flux, XY(), units = u"W/m^2", desc = "Latent heat flux at the surface [W m⁻²]"),
 )
 
+# The turbulent fluxes are prescribed input variables, so there is nothing to diagnose.
+@inline compute_auxiliary!(state, grid, ::PrescribedTurbulentFluxes, args...) = nothing
```
maximilian-gelbrecht, line 19: Ah so here it just dispatches on not computing them for the `TurbulentFluxes`. What actually happens when somebody wrongly configures the model with a `PrescribedSkinTemperature` but not with `PrescribedTurbulentFluxes`?

src/processes/surface/surface_energy_balance.jl
bgroenks96: Eh... it could, I guess, but it's not designed around a general `solve!` interface at the moment, so that would be arguably misleading. Also, this name is pre-existing to this PR.

src/processes/surface/turbulent_fluxes.jl
bgroenks96: At the moment, I think it would fail, but in principle, that should work, I think.

src/processes/surface/surface_energy_balance.jl
bgroenks96: This was a quick and dirty implementation, partly decided on by Claude. It circumvented the need to define more fallback dispatches at the process level. But looking at it again, I think I agree that it's probably better to add a bit more code to maintain a single execution path here in the fused kernel.

Done in [2fc7af7](https://github.com/NumericalEarth/Terrarium.jl/pull/155/commits/2fc7af72a244ad69f663566a2367f4c3f606968c).

src/processes/surface/surface_energy_balance.jl
maximilian-gelbrecht: Ok fine with me

maximilian-gelbrecht (approved):

PR #156, opened 2026-08-03
Add standard model performance benchmarks

maximilian-gelbrecht:
Modelled after SpeedyWeather's benchmarks. 

In short: benchmarks are run individually per architecture, results are stored in JSON files and are then build into one report including figures in the documentation. 

There's a list of model configurations that are run (similar to how it's done in the Reactant tests), but those are currently definitely not final yet. 

Otherwise: we scan resolution, number of layers, timestepper, precision.


maximilian-gelbrecht:
@bgroenks96 I am confused what you did with this PR. It says it's merged but I don't see it on `main`

bgroenks96:
I think GitHub closed it because your GPU PR included the commits.

PR #157, opened 2026-08-04
Reactant `LandModel` without vegetation

maximilian-gelbrecht:
This ensures Reactant compatibility of a `LandModel` without vegetation.


src/processes/soil/hydrology/soil_hydraulic_closures.jl
```diff
@@ -116,8 +116,10 @@ end
     por = porosity(i, j, k, grid, fields, strat, bgc)
     sat_res = residual_saturation(get_hydraulic_properties(hydrology))
     # compute matric pressure head
-    ψm = inv_swrc(max(sat * por, sat_res); θsat = por)
-    @assert isfinite(ψm) "NaN/infinite matric potential ψm = $ψm with sat=$sat por=$por resid=$sat_res"
+    # Note that `sat` is clamped from above as well as below: the inverse SWRC is evaluated with a
+    # non-negative base (see `VanGenuchten`/`BrooksCorey`), which oversaturated states (sat > 1) would
+    # violate, raising a `DomainError` from `^` — a throw path that cannot be compiled in a kernel.
+    ψm = inv_swrc(clamp(sat * por, sat_res, por); θsat = por)
```
maximilian-gelbrecht, line 119: @bgroenks96 I am a bit confused by this one. Claude did this automatically, do you think this is okay?

maximilian-gelbrecht: The `@assert` definitely has to be dropped though.

bgroenks96: It's fine, but remove or condense the verbose comment.

src/processes/soil/hydrology/soil_hydraulic_properties.jl
```diff
@@ -229,8 +229,11 @@ function hydraulic_conductivity(
         Ω = hydraulics.unsat_hydraulic_cond.impedance # scaling parameter for ice impedance
         I_ice = 10^(-Ω * (1 - f)) # ice impedance factor
         K_sat = saturated_hydraulic_conductivity(hydraulics, soil.solid.texture)
-        # We use `complex` types here to permit illegal state values which may occur when using adaptive time steppers.
-        K = abs(K_sat * I_ice * sqrt(complex(θw / θsat)) * (1 - complex(1 - complex(θw / θsat)^(n / (n + 1)))^((n - 1) / n))^2)
+        # Effective saturation, clamped to the unit interval so that every fractional power below is
+        # evaluated at a non-negative base. Illegal state values (θw > θsat or θw < 0), which may occur
+        # when using adaptive time steppers.
+        S = clamp(θw / θsat, zero(θw), one(θw))
+        K = K_sat * I_ice * sqrt(S) * (one(S) - (one(S) - S^(n / (n + 1)))^((n - 1) / n))^2
```
bgroenks96, line 240: Also fine.

src/processes/surface/skin_temperature.jl
```diff
@@ -80,7 +80,7 @@ Seed the prognostic `skin_temperature` with the current `ground_temperature` so
 nonlinear solve starts from a physically sensible guess close to the root.
 """
 function initialize!(state, grid, ::ImplicitSkinTemperature, args...)
-    set!(state.skin_temperature, state.ground_temperature)
+    interior(state.skin_temperature) .= interior(state.ground_temperature)   # Now set! for better Reactant compat, set! might make a CPU detour
```
bgroenks96: The comment doesn't match the change here. Why do we need to stop using `set!`? I would rather keep that.

bgroenks96 (dismissed):
Looks overall good, just look at the `set!` thing. That would be problematic if Reactant can't handle that.

src/processes/surface/skin_temperature.jl
maximilian-gelbrecht: It's just a typo "No set!". 

The problem is that `set!` might make CPU detours and the interpolation isn't fully Reactant compatible.

src/processes/soil/hydrology/soil_hydraulic_closures.jl
maximilian-gelbrecht: Removed

src/processes/surface/skin_temperature.jl
bgroenks96: Did you create an Oceananigans issue about that?

I don't see any reason why `set!` should make CPU detours when setting a `Field` to another `Field`.

maximilian-gelbrecht: No, not yet

bgroenks96: Ok, then please do that and link it here, and add a TODO comment in the code.

maximilian-gelbrecht:
Oh I thought this would already pass but it turns out there is another upstream Oceananigans Reactant issue with boundary conditions that we hit here. I converted this to draft for now. 

The main problem is the 2D indexing of `Fields` as it seems.  

I need to sit down and look at this properly to understand exactly what we need and maybe fix the issue upstream in Oceananigans.

src/solvers/newton.jl
```diff
@@ -0,0 +1,62 @@
...
+straight-line code.
+
+Unlike [`RootSolver`](@ref) and [`FixedPointSolver`](@ref) there is no convergence test and hence no
+data-dependent loop bound. That matters in two places:
+
+- **Reactant**: a convergence-tested loop lowers to an `scf.while` with a dynamic trip count, which the
+  StableHLO raise pass cannot lift. An unrolled loop raises cleanly.
+- **Reverse-mode AD**: a fixed trip count avoids a dynamically-sized tape.
+"""
+struct NewtonSolver{NF, iterations} end
```
bgroenks96, line 17: Why? We already had a Newton solver before and got rid of it in favor of RootSolvers.jl.

Can't you just disable the convergence test for the RootSolvers one?

bgroenks96: Also, this is again something that should be fixed with a custom rule... the AD system should not be differentiating through the iterations.

maximilian-gelbrecht: No custom rules with Reactant, but you are probably right with RootSolvers.

bgroenks96: Is there any documentation or roadmap anywhere about custom rules? I can't find anything, and I find this really concerning.

bgroenks96:
>  The main problem is the 2D indexing of Fields as it seems.

It would not be so objectionable to do `field[i, j, 1]` rather than `field[i, j]` everywhere. We already have to do this anyway when using `setindex!`.

src/processes/surface/skin_temperature.jl
maximilian-gelbrecht: This happens apparently because `skin_temperature` is `Field{Center, Center, Nothing}` and `ground_temperature` is `Field{Center, Center, Center}`. This triggers an interpolation then.

maximilian-gelbrecht:
> > The main problem is the 2D indexing of Fields as it seems.
> 
> It would not be so objectionable to do `field[i, j, 1]` rather than `field[i, j]` everywhere. We already have to do this anyway when using `setindex!`.

The culprit is in Oceananigans though https://github.com/CliMA/Oceananigans.jl/issues/5846

src/solvers/newton.jl
maximilian-gelbrecht: I am not terribly concerned around it as even the worst case of just differentiating through the maximum itertation number might not be that bad. But we are also not alone with this https://github.com/CliMA/Oceananigans.jl/issues/5844

src/processes/surface/skin_temperature.jl
bgroenks96: Ahh... I guess this would also happen if `skin_temperature` were defined at a `Face` (as it could be). 

The interpolation is probably pointless in this case unless the soil temperature has a Dirichlet boundary condition.

src/solvers/newton.jl
bgroenks96: An AD system without the ability to define gradients is not a complete AD system...

But anyway, I think we get around this by using `ignore_deriatives` which should be supported by Reactant, see https://github.com/EnzymeAD/Reactant.jl/pull/1373

maximilian-gelbrecht: RootSolvers isn't really compatible with Reactant as it is, it seems to me. Too many runtime conditions that can't be deactivated. I think I would keep this simple newton solver for the Reactant for now and then we might revisit this later depending on how we proceed with Reactant support (and the rest of the model)

bgroenks96: I would prefer to just do `ignore_derivatives` if possible?

maximilian-gelbrecht: I am not sure how `ignore_derivatives` should help here. This isn't only about taking derivatives. It's about running the model with Reactant also forward.

src/processes/surface/skin_temperature.jl
maximilian-gelbrecht: But the behaviour you intended for the initialization is to just copy the values, or do we actually want the interpolation in case of different locations?

src/solvers/newton.jl
bgroenks96: What precisely is the issue? It fails to compile entirely?

bgroenks96: It's things like this that make me not so convinced by Reactant. There seems to be a growing number of things that we now can't do in our code because of it, and that seems to be effectively just giving up on the original promise of "differentiate arbitrary Julia code". By the way, JAX absolutely can differentiate loops with tracked conditionals, see [jax.lax.while_loop](https://docs.jax.dev/en/latest/_autosummary/jax.lax.while_loop.html).

maximilian-gelbrecht: Reactant can also absolutely do that, but then the loop needs to be implemented like that e.g. with a `@trace` but RootSolvers.jl don't do that https://enzymead.github.io/Reactant.jl/stable/tutorials/control-flow#Loops

But it's also true that both for Reactant and for GPU in general these kind of data dependent kernels might just be undesirable performance wise as well because they may cause too many branch divergences.

bgroenks96: Then let's just make a PR to RootSolvers.jl to add `@trace`. ReactantCore is a lightweight package, and as long as it doesn't affect their code when *not* using Reactant, I don't see why they should object to it.

olivierbonte: First attempt at RootSolvers.jl compatible with Reactant (not yet complete) https://github.com/CliMA/RootSolvers.jl/pull/99 

The main problem that I am now at is that there is data-dependent while loop see https://github.com/olivierbonte/RootSolvers.jl/blob/db2a8349ecfaf9cae00516dac139286f52e930c8/src/RootSolvers.jl#L1851. As in the MWE found at https://github.com/CliMA/Oceananigans.jl/issues/5844, it is also here the case you you can make it work for the normal forward evaluation, but the AD fails. Following the suggestion of https://github.com/CliMA/Oceananigans.jl/issues/5844#issuecomment-5171628126 I could try to rewrite this.

So unfortunately not super straightforward to make it compatible. A rule system for AD would be nice, unfortunate that this is not (yet?) in Reactant.

maximilian-gelbrecht: That's already a great start Olivier! 

In the long run it would be nice if RootSolvers would be compatible, best with adjoint rules, but for the time being I am really not too worried. The Reactant model is already very very fast without any attempt to profile it (normal runs and AD), and if we have a separate fixed iteration version for Reactant for the time being that's not so bad and still quite performant.

maximilian-gelbrecht:
> > > The main problem is the 2D indexing of Fields as it seems.
> > 
> > 
> > It would not be so objectionable to do `field[i, j, 1]` rather than `field[i, j]` everywhere. We already have to do this anyway when using `setindex!`.
> 
> The culprit is in Oceananigans though [CliMA/Oceananigans.jl#5846](https://github.com/CliMA/Oceananigans.jl/issues/5846)

This issue is resolved (but not yet released). 

I'll pick up this PR when I find some time, maybe on the way home from JuliaCon.

src/solvers/newton.jl
bgroenks96: @olivierbonte I also attempted it (not sure if I pushed it to my fork) and ran into the same issues. I gave up after a couple of hours. I don't think it's entirely an issue of RootSolvers.jl being incompatible; in theory Reactant can handle data-dependent loops. But I hit a point where it looked like bugs in Reactant/StableHLO and just gave up.

olivierbonte: Alright, fixed point iteration for Reactant for now sounds reasonable!
Good to know you also gave it a shot @bgroenks96! I'll try to look at it a bit more if I find the time, but not sure If I will get the AD working. Do you think there is value in trying to get a PR merged in RootSolvers.jl that works for forward evaluations (but not yet for Enzyme)? Because that I think should work

bgroenks96: If the changes to the code are not too substantial, I would try it. But the CliMA project is pretty explicitly disinterested in AD so I doubt they will accept a major rewrite for Reactant.

maximilian-gelbrecht: If you can reduce it to a Reactant MWE if something isn't working that would be great as well though, we should absolutely post it in their repo. Usually responses are quick.

olivierbonte: https://github.com/CliMA/RootSolvers.jl/pull/99 now works for primal evaluations. Enzyme autodiff still WIP

src/utils/kernel_utils.jl
```diff
@@ -1,3 +1,18 @@
+"""
+    @assert_kernel cond [text]
```
maximilian-gelbrecht, line 18: I added `@assert_kernel` macro now that just disappears within a Reactant compile context, but otherwise just forwards to `Base.@assert`. This way we can still assert basically as before.

src/processes/surface/skin_temperature.jl
bgroenks96: I intended for it to just be copied.

maximilian-gelbrecht:
This works now. I would keep the fixed iteration `NewtonSolver` as a temporary solution for Reactant until we have something different figured out (like RootSolvers.jl compatability).

olivierbonte:
@maximilian-gelbrecht have you tried the RootSolvers.jl version at https://github.com/CliMA/RootSolvers.jl/pull/99? This was working okay for normal (non AD) reactant code evaluations but haven't tried it yet within Terrarium.

src/processes/surface/skin_temperature.jl
bgroenks96: Also, I still don't understand why interpolation would happen on CPU. That seems like a bug? @glwagner

glwagner: Hmm, it's true that different locations will trigger interpolation I think...

https://github.com/CliMA/Oceananigans.jl/blob/c8a95be84673affcefe59aa12592de7f6f88f60c/src/Fields/set!.jl#L164

and in particular here

https://github.com/CliMA/Oceananigans.jl/blob/c8a95be84673affcefe59aa12592de7f6f88f60c/src/Fields/set!.jl#L182-L183

we could perhaps fix this edge case by changing the condition `ℓu == ℓv` to a more specific condition that fails only for `Center` and `Face` but returns `true` if one of the locations is `Nothing`. I think that would allow this case to work right?

Something we were trying to prevent (which I believe did occur as a bug) is merely checking the size of the fields. As an example, `set!(u, v)` definitely should not behave like `interior(u) .= interior(v)` if `u` and `v` are at `Center` and `Face` in a periodic direction.

Unrelated, it also looks like there's a bug if the two fields are on the same size grid when those grids cover different spatial extents.

Also, we really do need Reactant support for interpolation. I think this is kind of important.

I don't believe that anything should go to CPU though. That should only happen if the target is on CPU.

bgroenks96: > we could perhaps fix this edge case by changing the condition ℓu == ℓv to a more specific condition that fails only for Center and Face but returns true if one of the locations is Nothing. I think that would allow this case to work right?

Doesn't it already do that with `reduced_singleton_dimension`? What is `N` here? The length of the dimension?

glwagner: Indeed, it should have. Except that reactant defines its own `set_to_field!`

So the somewhat funny situation is that the Reactant extension of `set_to_field!` actually causes `set!` to fail with Reactant, when it would otherwise succeed.

It's fixed in a new PR.

glwagner: Also on this:

> I don't believe that anything should go to CPU though. That should only happen if the target is on CPU.

we hit CPU only for interpolation (which isn't yet supported with Reactant), which is what we fall back to if we decide that broadcasting / copying will not work.

maximilian-gelbrecht: Thanks Greg! 

I reverted to `set!` and it's not yet working here end-to-end, I'll investigate a bit later.

maximilian-gelbrecht: Ah sorry, stupid me, I only added the branch in the main and normal test env of Terrarium and not in the Reactant test env. It does actually work.

maximilian-gelbrecht: Ah sorry, stupid me, I only added the branch in the main and normal test env of Terrarium and not in the Reactant test env. It does actually work.

maximilian-gelbrecht:
Further Oceananigans issues that blocked the vegetation model are addressed in https://github.com/CliMA/Oceananigans.jl/pull/5880

src/solvers/newton.jl
bgroenks96: I guess we will keep the Newton solver for now.

bgroenks96:
@maximilian-gelbrecht Is this ready to be merged?

maximilian-gelbrecht:
Yes with the new Oceananigans version released yesterday, we can merge this.

bgroenks96 (approved):

PR #160, opened 2026-08-06
Restore full GPU compat of `LandModel`

maximilian-gelbrecht:
Currently still doesn't work because of one of the same issues that also block #157 : https://github.com/CliMA/Oceananigans.jl/issues/5846


benchmark/assets/benchmark_results.json
bgroenks96, line 1: Um... I'm confused why this is here? isn't this branched off of `main`?

src/processes/soil/hydrology/soil_hydraulic_closures.jl
```diff
@@ -117,7 +117,6 @@ end
     sat_res = residual_saturation(get_hydraulic_properties(hydrology))
     # compute matric pressure head
     ψm = inv_swrc(max(sat * por, sat_res); θsat = por)
-    @assert isfinite(ψm) "NaN/infinite matric potential ψm = $ψm with sat=$sat por=$por resid=$sat_res"
```
bgroenks96, line 120: Could we maybe leave this here but commented out? It's a common issue that occurs so it's nice for diagnostics to have it already there.

src/solvers/fixed_point.jl
```diff
@@ -42,15 +42,15 @@ end
     # updated iterate is g(x) = x - F(x). Take the first (un-relaxed) step.
     residual = objective_func!(out, indices..., grid, fields, func_args...; func_kwargs...)
     x₁ = x₀ - residual
-    target_field[indices...] = x₁
+    target_field[field_indices(indices)...] = x₁
```
bgroenks96, line 45: Hmm, this seems a bit roundabout, but OK I guess.

I think I did at one point have a pirate override of `getindex`/`setindex!` for `Field`s that forwarded the 2-index to the corresponding 3-index call. We could consider re-adding that.

benchmark/assets/benchmark_results.json
bgroenks96: maybe mixed up commits with #156 ?

src/processes/soil/hydrology/soil_hydraulic_closures.jl
maximilian-gelbrecht: I put the assert back in just without interpolated strings. In the Reactant version we might have to comment it out though.

benchmark/assets/benchmark_results.json
maximilian-gelbrecht: I started this from benchmark branch but removed the benchmarks, I am confused by the comment though because this file isn't there anymore.

src/solvers/fixed_point.jl
maximilian-gelbrecht: Do you think this could cause trouble when coupling to Oceananigans in the end or should that be fine?

bgroenks96: Hmm... good point. I guess this is safer.

benchmark/assets/benchmark_results.json
bgroenks96: I meant why is the change there.

maximilian-gelbrecht: I mean, it's not here anymore and also wasn't when I opened the PR that's why I am confused as well. I am going to squash the commits in the end on this PR.

bgroenks96 (approved):

maximilian-gelbrecht:
I'll wait for https://github.com/CliMA/Oceananigans.jl/pull/5849 to merge this (and properly test the GPU model)

bgroenks96:
Verified locally on `land_column.jl` example.

Issue #107, opened 2026-04-20
Slow CPU performance global run


olivierbonte:
When following the [global soil heat example](https://github.com/NumericalEarth/Terrarium.jl/blob/ca59c6018a7d993032b665b55a2cfa25d6cbe787/examples/simulations/soil_heat_global.jl), I noticed a very large difference between CPU and GPU performance, which might be of interest. 

Executing
https://github.com/NumericalEarth/Terrarium.jl/blob/ca59c6018a7d993032b665b55a2cfa25d6cbe787/examples/simulations/soil_heat_global.jl#L101
on GPU gives:
```julia 
0.028336 seconds (106.91 k allocations: 19.431 MiB)
```
on CPU on the other hand:
```julia
74.107660 seconds (578.58 M allocations: 223.602 GiB, 31.80% gc time)
```
which is quite a substantial difference

Some more information on the sytstem
```julia
Julia Version 1.11.9
Commit 53a02c0720c (2026-02-06 00:27 UTC)
Build Info:
  Official https://julialang.org/ release
Platform Info:
  OS: Linux (x86_64-linux-gnu)
  CPU: 8 × 11th Gen Intel(R) Core(TM) i7-1165G7 @ 2.80GHz
  WORD_SIZE: 64
  LLVM: libLLVM-16.0.6 (ORCJIT, tigerlake)
Threads: 1 default, 0 interactive, 1 GC (on 8 virtual cores)
Environment:
  JULIA_EDITOR = code
  JULIA_VSCODE_REPL = 1
```
packages 
```julia
Status `~/Doctoraat/Terrarium.jl/examples/Project.toml`
⌃ [052768ef] CUDA v5.11.0
  [13f3f980] CairoMakie v0.15.9
⌃ [eb46d486] Checkpointing v0.11.2
  [0b91fe84] DisplayAs v0.1.6
  [7da242da] Enzyme v0.13.138
  [db073c08] GeoMakie v0.7.16
  [a98d9a8b] Interpolations v0.16.2
  [033835bb] JLD2 v0.6.4
  [63c18a36] KernelAbstractions v0.9.41
  [85f8d34a] NCDatasets v0.14.15
⌅ [9e8cae18] Oceananigans v0.106.6
  [a3a2b9e3] Rasters v0.14.7
  [d1845624] RingGrids v0.1.3
  [9e226e20] SpeedyWeather v0.18.1
  [34489162] SpeedyWeatherInternals v0.1.3
[...]

PR #163, opened 2026-08-10
Add performance benchmarks

maximilian-gelbrecht:
Supersedes #156 

> Modelled after SpeedyWeather's benchmarks.
> 
> In short: benchmarks are run individually per architecture, results are stored in JSON files and are then build into one report including figures in the documentation.
> 
> There's a list of model configurations that are run (similar to how it's done in the Reactant tests), but those are currently definitely not final yet.
> 
> Otherwise: we scan resolution, number of layers, timestepper, precision.

With the respect to the results, we see here a massive difference between the non-Reactant and Reactant versions so far.  We have to profile this to be sure, but my guess is that the non-Reactant version has a ton of overhead (e.g from kernel launching) that is eliminated by Reactant for which kernel launching overheads don't exist anymore. The difference is so massive that this really raises the question whether we want to change something about the design there or whether we just regard Reactant as the performance-default in the future. But again, we have to profile this better to know for sure.


maximilian-gelbrecht:
Docs preview: https://numericalearth.github.io/Terrarium.jl/previews/PR163/benchmarks/ 

There are certainly also a few outliers from the expected monotonic behaviour in there. This should probably mean that the default number of time steps is not large enough.

bgroenks96:
The kernel launch overhead is a known issue, discussed a bit in #107, but doesn't currently have its own issue (which it should). The pending fix for this is to use fused kernels for soil and vegetation that mirror the structure used already by the `SurfaceEnergyBalance`. I just haven't gotten around to doing it yet.

I would expect that, once these changes are made, that gap will narrow considerably.

maximilian-gelbrecht:
@bgroenks96 If you have some time, you take a look at it.

benchmark/manual_benchmarking.jl
```diff
@@ -0,0 +1,289 @@
+#=
+Run the Terrarium benchmark suite on one architecture and merge the result into a shared `README.md`
```
bgroenks96, line 2: Could you please ask Claude to refactor this to use `ArgParse` so that this is documented by the command rather than in a comment?

bgroenks96 (changes requested):
Looks generally good, I made a couple of small comments. I won't look at the code in great detail since it's non-critical, but I already looked at the doc pages and liked the overall structure + formatting.

bgroenks96 (changes requested):
Looks generally good, I made a couple of small comments. I won't look at the code in great detail since it's non-critical, but I already looked at the doc pages and liked the overall structure + formatting.

AGENTS.md
```diff
@@ -195,6 +195,32 @@ src/
 14. **Mutable function closures in kernels**: function closures that capture mutable state will not differentiate correctly — use explicit parameters instead
 15. **Non-local dependencies in process equations**: process functions must not depend on global state; pass all dependencies as arguments for traceability and differentiability
 
+## Benchmarks
+
+Performance benchmarks live in `benchmark/` and are run manually, one architecture at a time:
```
bgroenks96, line 200: I guess we don't want Claude to run these every time it makes a change, right? Should we add something about only doing it when instructed?

maximilian-gelbrecht: We also have this in Speedy's CLAUDE.md, this is where this is copied from, and it only ever ran them automatically on really major revisions (where it also makes sense). However, adding something like this explicitly also makes sense.

benchmark/manual_benchmarking.jl
maximilian-gelbrecht: Done

bgroenks96 (approved):

PR #165, opened 2026-08-10
Fix bugs in snow and canopy ET + add global land simulation example

bgroenks96:
This PR adds a new global `LandModel` simulation example script that exercises snow + prescribed LAI in an offline ERA5-Land forced simulation. It also updates the SpeedyWeather `PrimitiveWetModel` coupling script to include snow. Last time I checked, the ERA5 simulation wasn't working due to instabilities, but that may be fixed now, still need to check.

## Summary of changes
- Fix bugs and numerical stability issues in single-layer snow scheme
- Fix bugs in canopy ET dispatch pattern
- Update ERA5-Land LAI product
- Add optional `ProgressMeter` support in `run!` (currently opt-in with `show_progress = true`)

## TODO
- [x] Check global land example again


Project.toml
```diff
@@ -1,7 +1,7 @@
 name = "Terrarium"
 uuid = "80418d68-07fa-499d-ae2b-c12e531f5cd8"
 authors = ["Brian Groenke <brian.groenke@pik-potsdam.de>", "Maximillian Gelbrecht <gelbrecht@pik-potsdam.de>", "Maha Badri <maha.badri@pik-potsdam.de>", "Terrarium Contributors"]
-version = "0.1.4"
+version = "0.1.5+DEV"
```
maximilian-gelbrecht: I would prefer to do version bumps and releases in separate PRs, also why directly `+DEV`?

examples/simulations/land_global_era5.jl
```diff
@@ -0,0 +1,192 @@
...
+# downwelling radiation) is supplied as time-varying inputs from the [`ERA5LandForcings`](@ref)
+# asset via the [`InputSource`](@ref) interface, which linearly interpolates each field in time.
+#
+# !!! warning "Illustrative configuration"
+#     This is a demonstration of the coupled-model setup, not a validated land-surface product. The
+#     ERA5-Land accumulated fluxes (radiation and precipitation) are de-accumulated to hourly rates
+#     as described below; the only approximation is that the small amount of flux falling exactly on
+#     the nightly accumulation reset is discarded. The script is intended to be run directly and is
+#     *not* executed during the documentation build, as it downloads several gigabytes of forcing
+#     data and integrates the full global grid.
```
maximilian-gelbrecht, line 22: Besides this PR: But I am wondering whether we eventually want some kind of system like NumericalEarth.jl that occasionally but not always builds these scripts so that they can still be displayed in the docs, same also goes for the differentiability scripts.

examples/simulations/land_global_era5.jl
```diff
@@ -0,0 +1,192 @@
...
+# dewpoint temperature and surface pressure with [`Terrarium.dewpoint_specific_humidity`](@ref),
+# using a standalone set of thermodynamic constants.
+air_temperature = t2m .- NF(273.15)   # K -> °C
+air_pressure = sp                     # Pa
+wind_u = u10                          # m/s
+wind_v = v10                          # m/s
+constants = ThermodynamicConstants(NF)
+specific_humidity = Terrarium.dewpoint_specific_humidity.(Ref(constants), d2m .- NF(273.15), sp)
+
+# The *accumulated* fields (radiation and precipitation) are cumulative since 00 UTC and reset each
```
maximilian-gelbrecht, line 90: Interesting! I did not know that.

src/diagnostics/debugging.jl
```diff
@@ -1,15 +1,4 @@
-global DEBUG::Bool = haskey(ENV, "TERRARIUM_DEBUG") && ENV["TERRARIUM_DEBUG"] == "true"
-
-"""
-    debug!(debug::Bool)
-
-Enable or disable global debug mode for Terrarium. Debug mode 
-"""
-function debug!(debug::Bool)
```
maximilian-gelbrecht, line 8: Completely removed or just moved/replaced?

src/models/coupled/land_model.jl
```diff
@@ -139,6 +139,17 @@ function invclosure!(state, model::LandModel)
     return nothing
 end
 
+# Time time_stepping
+
+function Terrarium.timestep!(state::StateVariables, model::LandModel, timestepper::Terrarium.AbstractTimeStepper, Δt)
+    if !isnothing(model.snow)
```
maximilian-gelbrecht, line 145: I would rewrite this to just call `timestep!(state, model.snow, timestepper, Δt)` (or even more general call `timestep(state, model, ...)` on all model components). 

With a fallback `timestep!(state, ::Union{AbstractProcess,Nothing}, timestepper, Δt) = nothing`, you wouldn't even need the `if` here

maximilian-gelbrecht (commented):
Looks good!

Project.toml
bgroenks96: in my mind, the `DEV` version should be for the version not yet released, not for the one already released. But maybe I am misunderstanding something about Julia version semantics.

examples/simulations/land_global_era5.jl
bgroenks96: Yes that would be nice.

examples/simulations/land_global_era5.jl
bgroenks96: I think I did but forgot about it until Claude told me.

src/diagnostics/debugging.jl
bgroenks96: Completely removed.

Project.toml
maximilian-gelbrecht: This would be `0.1.5-DEV` if it's the not yet released version

src/models/coupled/land_model.jl
bgroenks96: I agree with this as a more general pattern, but I also tend towards keeping things simple until the need for that general pattern actually becomes present. So in this case, where we only need to do something special for snow, it seems more straightforward to to do this..?

Also this shouldn't affect Reactant since it's a static conditional.

examples/simulations/land_global_era5.jl
maximilian-gelbrecht: I just opened an issue, so we don't forget https://github.com/NumericalEarth/Terrarium.jl/issues/166

maximilian-gelbrecht: I just opened an issue, so we don't forget https://github.com/NumericalEarth/Terrarium.jl/issues/166

src/models/coupled/land_model.jl
maximilian-gelbrecht: I just don't like this code duplication here. We don't have to the very general pattern, but at least just doing 

```
function Terrarium.timestep!(state::StateVariables, model::LandModel, timestepper::Terrarium.AbstractTimeStepper, Δt)
    if !isnothing(model.snow)
         timestep!(state, model.snow, timestepper, Δt) # clamp snow to be non-negative
    end 
end 
```
would reduce a bit of duplication.

src/processes/thermodynamics/thermodynamics.jl
```diff
@@ -71,7 +71,9 @@ pressure is computed over ice for `T <= 0°C` and over water for `T > 0°C`. Wra
 [`saturation_vapor_pressure`](@extref Thermodynamics.saturation_vapor_pressure).
 """
 @inline function saturation_vapor_pressure(c::ThermodynamicConstants, T::NF) where {NF}
-    T_K = celsius_to_kelvin(c, T)
+    # Floor the absolute temperature at `eps` before entering Thermodynamics.jl: its Clausius–Clapeyron
+    # form evaluates `log(T_K / T_triple)`, which throws a `DomainError` for `T_K ≤ 0`
+    T_K = max(celsius_to_kelvin(c, T), eps(NF))
```
bgroenks96, line 69: @olivierbonte take note of this

src/models/coupled/land_model.jl
bgroenks96: That doesn't work in the current code because `model.snow` is not a `SnowModel`. But I'll see what I can do.

bgroenks96: Fixed in [f3719fe](https://github.com/NumericalEarth/Terrarium.jl/pull/165/commits/f3719fe73247f41427e2b841ed6fed8842a4d5b1)

Project.toml
bgroenks96: Fixed in [eb19656](https://github.com/NumericalEarth/Terrarium.jl/pull/165/commits/eb196562c8e209f3286d1da725f7934264557da1)

maximilian-gelbrecht (approved):

examples/simulations/land_global_era5.jl
```diff
@@ -0,0 +1,192 @@
...
+# [`ColumnRingGrid`](@ref) at ~1° resolution (72 Gaussian rings), keeping only grid points with more
+# than 50% land cover. The land-sea mask is loaded from the ERA5-Land invariants asset and regridded
+# onto the model grid. We use an exponentially-spaced soil column so that the near-surface layers,
+# where the diurnal and seasonal signals are strongest, are resolved most finely.
+model_rings = on_architecture(arch, RingGrids.FullGaussianGrid(72))
+land_sea_frac_native = RingGrids.Field(arch, ERA5LandInvariants(), "lsm"; NF)
+land_sea_frac = RingGrids.interpolate(model_rings, land_sea_frac_native)
+land_mask = land_sea_frac .> 0.5
+Nz = 30 # number of soil layers
+grid = ColumnRingGrid(arch, NF, ExponentialSpacing(N = Nz, Δz_min = 0.05, Δz_max = 2.0), land_mask)
```
olivierbonte: More of a question than a comment: is it best practice to use `NF(0.05)` and `NF(2.0)` here? Or is this not necessary when you set NF for the grid type?

examples/simulations/land_global_era5.jl
```diff
@@ -0,0 +1,192 @@
...
+# The model consumes inputs in its own units, so we apply the necessary conversions as lazy
+# broadcasts over the rasters. Only the slices touched by the time interpolation are ever evaluated,
+# and because the ERA5-Land fields are already stored in `Float32` these single-level arithmetic
+# broadcasts preserve the numeric type expected by the model grid.
+#
+# The *instantaneous* fields (temperature, dewpoint, pressure, wind) need only unit conversions.
+# Near-surface specific humidity is not provided directly by ERA5-Land, so we derive it from the
+# dewpoint temperature and surface pressure with [`Terrarium.dewpoint_specific_humidity`](@ref),
+# using a standalone set of thermodynamic constants.
+air_temperature = t2m .- NF(273.15)   # K -> °C
```
olivierbonte: Consider using https://numericalearth.github.io/Terrarium.jl/v0/processes/utils/physical_constants/#Terrarium.celsius_to_kelvin-processes-utils-physical_constants for this?

examples/simulations/land_global_era5.jl
```diff
@@ -0,0 +1,192 @@
...
+
+# ## Initial conditions
+# The soil moisture starts from a variably-saturated profile that approaches saturation with depth,
+# and the surface starts snow-free so the winter forcing builds up the snowpack. The soil temperature
+# is anchored to the local ERA5 near-surface air temperature: we initialize the model once (which
+# loads the forcings at `t = 0`), read the resulting per-column air temperature from the state, and
+# re-initialize with a soil temperature profile built from it plus a mild geothermal gradient. Reading
+# the air temperature back from the state guarantees it is ordered consistently with the land columns.
+initializers = (
+    # saturation_water_ice = (x, z) -> min(one(NF), NF(0.6) - NF(0.05) * z),
```
olivierbonte: Why is this commented out? Isn't this the saturation profile we want?

examples/simulations/land_global_era5.jl
```diff
@@ -0,0 +1,192 @@
...
+
+fig = plot_surface(integrator.state.leaf_area_index, title = "Leaf area index (Jan)", colorrange = (0, 6))
+DisplayAs.PNG(fig) #hide
+
+# ## Run through the first three months
+# We advance the coupled model for 2 days with a 15-minute step. The explicit `ForwardEuler`
+# stepper requires a step size small enough to keep the soil dynamics stable.
+Terrarium.initialize!(integrator)
+@profview @time timestep!(integrator)
+run!(integrator, period = Minute(30), Δt = Minute(1))
```
olivierbonte: This does not align with the description: 30 minute <-> 2 days, 1 minute <-> 15 minutes

examples/simulations/land_global_era5.jl
```diff
@@ -0,0 +1,192 @@
...
+
+# The surface soil temperature and accumulated snow water equivalent after three months of forcing.
```
olivierbonte: Integration is not for 3 months

examples/simulations/speedy_wet_land.jl
```diff
@@ -14,30 +17,76 @@ import SpeedyWeather as Speedy
...
+# the poles.
+mean_annual_temperature(lat) = 20 - abs(40 * sin(lat)) # maximum at equator
+
+function initial_soil_temperature(grid)
+    arch = Terrarium.architecture(grid)
+    _, grid_lat = RingGrids.get_lonlats(grid.rings) # in radians
+    function init(x, z)
+        latᵢ = grid_lat[round(Int, x)]
+        T₀ = mean_annual_temperature(latᵢ)
+        T = T₀ - 0.05 * z
```
olivierbonte, line 40: NF needed for 0.05?

examples/simulations/land_global_era5.jl
```diff
@@ -0,0 +1,192 @@
...
+specific_humidity = Terrarium.dewpoint_specific_humidity.(Ref(constants), d2m .- NF(273.15), sp)
+
+# The *accumulated* fields (radiation and precipitation) are cumulative since 00 UTC and reset each
+# day, so dividing the raw values by the output interval would badly overestimate the flux. We
+# instead recover the instantaneous hourly rate as the positive time difference between consecutive
+# records divided by the accumulation window; `max(·, 0)` discards the daily reset, which falls at
+# night where the radiative and (typically) precipitation fluxes are negligible. `deaccumulate`
+# returns a lazy raster spanning the records from the second hour onward.
+accumulation_window = NF(3600) # seconds
+function deaccumulate(accumulated)
```
olivierbonte: My laptop crashes (out of RAM) when trying to apply this deaccumulate. This makes me presume that these calculations are potentially not lazy?

src/processes/surface/evapotranspiration/canopy_evapotranspiration.jl
```diff
@@ -138,14 +141,14 @@ of the skin temperature.
 
     # Compute aerodynamic resistances and resistance factors
     rₐ = aerodynamic_resistance(i, j, grid, fields, atmos) # aerodynamic resistance
-    rₑ = aerodynamic_resistance(i, j, grid, fields, atmos, evapotranspiration, vegetation) # aerodynamic resistance between ground and canopy
+    rₛ = aerodynamic_resistance(i, j, grid, fields, atmos, evapotranspiration, vegetation) # aerodynamic resistance between ground and canopy
```
olivierbonte: I wouldn't call this r_s as it is a not a surface resistance, but an extra aerodynamic resistance. so r_ae or something is more suited. Would revert to change r_s here in at other locations in the code

test/surface/evapotranspiration/evapotranspiration_tests.jl
```diff
@@ -155,13 +155,14 @@ end
...
-# Integration tests - full ET workflows
+# Integration tests
 # ==============================================================================
 
-@testset "Full canopy ET workflow" begin
+@testset "Canopy ET flux" begin
     canopy_ET = PALADYNCanopyEvapotranspiration(Float64)
 
     rₐ = 100.0
+    rₛ = zero(rₐ) # stomatal resistance
```
olivierbonte: This is not stomatal resistance -> remove comment

olivierbonte: You can set an example value here for this excess resistance, it doesn't have to be zero

olivierbonte (commented):
Looks good to me in general. I could not run the example locally (RAM issues it seems), do you have an idea of what the hardware requirements are?

examples/simulations/land_global_era5.jl
bgroenks96: In this case, it's not necessary.

examples/simulations/land_global_era5.jl
bgroenks96: I disabled soil hydrology for now.

examples/simulations/land_global_era5.jl
bgroenks96: Yes, I need to update the comments (or the code... depends on whether it runs).

examples/simulations/speedy_wet_land.jl
bgroenks96: I am actually not sure if this is necessary here, but I guess better to be safe/consistent.

examples/simulations/land_global_era5.jl
bgroenks96: Hmm, I think they are supposed to be, but it's possible that somewhere the arrays are getting loaded into memory. I'll check.

examples/simulations/land_global_era5.jl
olivierbonte: Okay good to know!

examples/simulations/land_global_era5.jl
olivierbonte: Ahn okay, in that case it would be good to change some of the descriptive text, as this now explicitly mentions solving the Richard Equation

bgroenks96: Yeah... I need to update all of the text.

test/surface/evapotranspiration/evapotranspiration_tests.jl
bgroenks96: Fixed in [bd51b47](https://github.com/NumericalEarth/Terrarium.jl/pull/165/commits/bd51b470e4b226c493e8d42a2eb79f8621bbc8bf) and [7b289e6](https://github.com/NumericalEarth/Terrarium.jl/pull/165/commits/7b289e66345f7d4428fabfd8fef5f009f978ab6f)

src/processes/surface/evapotranspiration/canopy_evapotranspiration.jl
bgroenks96: Fixed in [bd51b47](https://github.com/NumericalEarth/Terrarium.jl/pull/165/commits/bd51b470e4b226c493e8d42a2eb79f8621bbc8bf)

examples/simulations/land_global_era5.jl
bgroenks96: Interestingly, there wasn't yet a `kelvin_to_celsius` method.

bgroenks96:
Unfortunately, the ERA5-Land forced simulation is still running into NaNs and I haven't found the cause yet.

bgroenks96:
@olivierbonte Can you try running the example with [cad6ec3](https://github.com/NumericalEarth/Terrarium.jl/pull/165/commits/cad6ec374975720ff768cc444ab62c3835929daa) to see if it fixes your out-of-memory issue?

olivierbonte:
> @olivierbonte Can you try running the example with [cad6ec3](https://github.com/NumericalEarth/Terrarium.jl/pull/165/commits/cad6ec374975720ff768cc444ab62c3835929daa) to see if it fixes your out-of-memory issue?

I'll give it a shot!

bgroenks96:
@olivierbonte You'll need to pull again, the previous change to debug mode broke precompilation of the extensions.

olivierbonte:
It is currently not running because there is a bug in the temperature conversion: `kelvin_to_celsiuis` required a thermodynamic constant to be passed

bgroenks96:
Yeah sorry my bad, fixed now.

bgroenks96:
With the help of debug mode, I have narrowed it down to the plant available water somehow becoming NaN... which seems weird considering that soil moisture is constant.

bgroenks96:
Ok here is a problem:

The SURFEX parameterizations don't seem to handle the case where clay content is zero (defaults used here)

```julia
inline function wilting_point(hydraulics::SoilHydraulicsSURFEX, texture::SoilTexture)
    β_w = hydraulics.wilting_point_effect
    wp = β_w * sqrt(texture.clay * 100)
    return wp
end

@inline function field_capacity(hydraulics::SoilHydraulicsSURFEX, texture::SoilTexture)
    η = hydraulics.field_capacity_exp
    β_c = hydraulics.field_capacity_effect
    fc = β_c * (texture.clay * 100)^η
    return fc
end
```

because both the field capacity and wilting point are then zero and so the denominator for PAW is zero. Do you have any thoughts as to a suitable solution to that @olivierbonte?

bgroenks96:
For now I added a minimum field capacity parameter in [da79fe8](https://github.com/NumericalEarth/Terrarium.jl/pull/165/commits/da79fe8c89b51684818754fc5bed9053b7100f4f)

olivierbonte:
I think that is fine for now, in the long term, it would be good to move away from the SURFEX parametrisations. They are nice as backup, but the data they are fitted on is quite limited. SoilGrids also publish volumetric water contents (see https://doi.org/10.1016/j.iswcr.2022.08.001), so this could be a good source. There are many alternatives though of datasets that directly provide the soil water characteristics. E.g. hihydrosoil (https://gee-community-catalog.org/projects/hihydro_soil/#citation-related-publications) or https://doi.org/10.5194/essd-9-529-2017 are themselves based on soilgrids

olivierbonte:
with `@profview` deleted, this runs for me (CPU, WSL)!

bgroenks96:
I don't think this is going to work without timestepping improvements. I would suggest we just merge it as-is (running but not converging) and then address this in follow-up PRs.

bgroenks96:
To be continued in #167

examples/simulations/speedy_wet_land.jl
bgroenks96: Fixed now

bgroenks96:
@olivierbonte Can you take a quick look at the new test failure? It seems like maybe an update to Thermodynamics.jl? Because nothing really changed in the last commit except a comment.

bgroenks96:
Temporarily fixed in [5179c6f](https://github.com/NumericalEarth/Terrarium.jl/pull/165/commits/5179c6f627e7a52c4925d41d190f05b7a79cc03b). The issue is caused by Thermodynamics 1.3 which was released earlier today. Their CI is failing too so I guess they broke something.

bgroenks96:
Ok, for some reason GitHub CI is completely ignoring the compat bounds on Thermodynamics.jl. Any ideas @olivierbonte?

olivierbonte:
Seems like something simple with the element type of the constans, looking into it

bgroenks96:
Deleting the caches also didn't help. I can't reproduce this locally with the same Julia version (v1.10.11) with `import Pkg; Pkg.test(;coverage=true, julia_args=["--check-bounds=yes", "--compiled-modules=yes", "--depwarn=yes"], force_latest_compatible_version=false, allow_reresolve=true)` so this appears to be a GitHub issue, not a Julia issue. I am therefore going to bypass the merge rules and we can address this (hopefully) in a follow up PR.

PR #167, opened 2026-08-12
Add adaptive timestepping with CFL condition via `Simulation` and `TimeStepWizard`

bgroenks96:
This PR adds adaptive timestepping via the Oceananigans `TimeStepWizard`. An example of its use can be seen in the `land_global_era5.jl` script. The original motivation was to try to fix some stability issues there; however, this has not yet been successful. It seems the stability issues are more related to the prescribed vegetation and surface energy balance, not the soil.

Nevertheless, it's probably good to have this feature. Should we update all of the existing examples to use `Simulation` with `TimeStepWizard` instead of `integrator` directly?


docs/src/running/time_stepping.md
```diff
@@ -173,3 +173,65 @@ sim.callbacks[:progress] = Callback(print_progress, IterationInterval(100))
 
 The callback function receives the `Simulation` object, giving it access to the full
 `integrator` via `sim.model`, and to the current `state` via `sim.model.state`.
+
+## Adaptive time stepping
+
+The explicit time steppers integrate the (stiff) diffusive soil operators, so their stability is
+constrained by the diffusive Courant–Friedrichs–Lewy (CFL) condition
```
olivierbonte, line 180: Just a small comment on the term diffusive CFL: I see that the term is also used in Oceananigans, but I suppose this is equivalent to time step limit following van Von Neumann stability analysis (see e.g. [here](https://en.wikipedia.org/wiki/Von_Neumann_stability_analysis#Illustration_of_the_method)? I hadn't heard of CFL in the context of Parabolic PDEs, so might be good to make this link explicit in the docs

docs/src/running/time_stepping.md
```diff
@@ -173,3 +173,65 @@ sim.callbacks[:progress] = Callback(print_progress, IterationInterval(100))
...
+adaptive_sim = Simulation(adaptive_integrator; stop_iteration = 50, Δt = 60.0)
+
+wizard = TimeStepWizard(diffusive_cfl = 0.5, max_change = 1.1)
+adaptive_sim.callbacks[:wizard] = Callback(wizard, IterationInterval(5))
+
+run!(adaptive_sim)
+adaptive_sim.Δt   # adapted time step (s)
+```
+
+Land models transport no quantity advectively, so the companion advective diagnostic
```
olivierbonte, line 211: Doesn't the Richards equation have a advective component in the gravity transport (see e.g. See e.g. https://cloud.env.nm.gov/resources/_translator.php/Y2Q1ZDRmOWMxMTRiNGU5NDM4Yzc1NDUzY182NjMwMg~~.pdf)? I guess you could calculate an advective CFL for that

docs/src/running/time_stepping.md
bgroenks96: I'm not entirely sure about the historical connection between CFL and von Neumann stability; mathematically, they are indeed equivalent for diffusion, but I think you can arrive at the same answer here also starting from the CFL.

The CFL places a limit on the wave propagation speed based on the grid cell length $\Delta x$,
```math
\frac{ \lvert v \rvert \Delta t}{\Delta x} \leq C\,.
```
While diffusive systems do not have an independent velocity $v$, they do have a characteristic length scale based on the diffusion coefficient $\alpha$,
```math
\ell \sim \sqrt{\alpha \Delta t}
```
which we can understand as the distance over which information gets dispersed in a single time step $\Delta t$. Plugging that in as the numerator in CFL yields,
```math
\frac{\sqrt{\alpha \Delta t}}{\Delta x} \leq C
```
and solving for $\Delta t$ we get,
```math
\Delta t \leq \frac{C^2 \Delta x^2}{\alpha}\,.
```
Since $C$ is a constant, we can simply define $C_{\text{diff}} = C^2$; but the choice of $C$ is determined by von Neumann stability analysis.

docs/src/running/time_stepping.md
bgroenks96: Yes, I suppose we could, but I will need to double check this.

bgroenks96: I also don't like the wording "land models transport no quantity advectively"; that's clearly not true as a general statement.

docs/src/running/time_stepping.md
olivierbonte: That makes sense (and more intuitive than the von Neumann derivation IMO)! As you say, to explain why $C_{diff}$ is 1/2 for the explicit case, mentioning von Neumann is more logical

docs/src/running/time_stepping.md
bgroenks96: I think we can add the advection limit once we add heat advection to the soil energy-water coupling.

The docs are updated in [a400cee](https://github.com/NumericalEarth/Terrarium.jl/pull/167/commits/a400cee9c199493f35158a52d570d8ff991bf795).

docs/src/running/time_stepping.md
bgroenks96: Added to docs in [a400cee](https://github.com/NumericalEarth/Terrarium.jl/pull/167/commits/a400cee9c199493f35158a52d570d8ff991bf795) and [db80fa0](https://github.com/NumericalEarth/Terrarium.jl/pull/167/commits/db80fa092bfa4c9420577c43e3b7f01a31568aea).

docs/src/running/time_stepping.md
```diff
@@ -95,6 +95,104 @@ using Oceananigans.Units: days
...
+!!! warning "Richards equation is stiff near saturation"
+    As the soil saturates, ``\partial\theta/\partial\psi \to 0`` and the hydraulic timescale tends to
+    zero. Explicit integration of the Richardson–Richards equation is therefore extremely stiff in
+    the saturated limit, which is the regime that motivates implicit time stepping (planned). The
+    `TimeStepWizard`'s `min_change` bound prevents `Δt` from collapsing to zero in a single
+    adjustment, but a saturated explicit run may still require very small steps.
+
+!!! note "Reactant"
+    Adaptive stepping applies to host-driven `Simulation`s only. Under a `ReactantState` grid the
+    time step is compiled into the traced `run!` loop, so a host-side wizard callback does not take
```
maximilian-gelbrecht, line 186: Did you check that? I don't know, but as we discussed yesterday that would be easy to avoid in principle, I just don't know if that's actually implemented like that.

src/diagnostics/progress.jl
```diff
@@ -0,0 +1,44 @@
...
+    end
+end
+
+function (callback::ProgressReporter)(simulation)
+    desc = "Running simulation (Δt = $(simulation.Δt))"
+    if isfinite(simulation.stop_iteration)
+        return ProgressMeter.update!(callback.progress, simulation.model.clock.iteration; desc)
+    else
+        val = Int(ceil(simulation.model.clock.time / simulation.stop_time * 100))
+        return ProgressMeter.update!(callback.progress, val; desc)
```
maximilian-gelbrecht, line 37: Is this called at every time step? If so that's probably a bad idea for GPU simulations. There it would be good to only report progress every 50-100 steps or so.

maximilian-gelbrecht (approved):
Looks good to me just two minor questions.

docs/src/running/time_stepping.md
bgroenks96: The current implementation would be bypassed by Reactant, yes.

src/diagnostics/progress.jl
bgroenks96: I don't think this would execute on the GPU?

bgroenks96: As clarified verbally, there is no NaN check or device-to-host synchronization here. The progress reporting should be entirely host side.

PR #170, opened 2026-08-15
Fix stability issues in the coupling of `LandModel` (soil + snow + canopy)

bgroenks96:
This PR is a collection of fixes that were necessary to get the coupling of the full `LandModel` with `PrescribedVegetation` and `SingleLayerSnow` semi-working with both SpeedyWeather and ERA5-Land forcing.

## Summary of changes
- Rewrite the implementation of `MedlynStomatalConductance` to (a) be more physically consistent, and (b) handle corner cases like where the VPD is zero and `An < 0` (@mahabadri)
- Partial fix for `CanopyWaterInterception` accumulating negative water storage; clamping of the water storage term in each timestep should be added in a later PR
- Fix `PALADYNCanopyEvapotranspiration` allowing for negative transpiration; the transpirative humidity flux is now clamped at zero, but deposition may still occur via the canopy term
- Fix incorrect implementation of `compute_transpiration_conductance` that allowed for nonzero transpiration over unvegetated grid cells
- Fix incorrect handling of residual water content in soil moisture evaporation resistance factor
- Add a lower resolution (N72) variant of the ERA5-Land LAI climatology for faster data loading in examples
- Increase numerical precision of `R_v` and `R_d` in Thermodynamics constants to increase accuracy of derived universal gas constant
- Add more rigorous tests for SEB solver under snow-free, snow-covered, and vegetated conditions


docs/dev/2026-07/2026-07-08-PLAN_reactant.md
maximilian-gelbrecht, line 1: Not sure, I'd really want older dev files to be adjusted to new file names. 

Maybe add a rule to not touch plans that are marked as completed?

examples/simulations/speedy_wet_land.jl
```diff
@@ -116,33 +141,44 @@ Speedy.add!(primitive_wet_coupled.output, Speedy.AlbedoOutput())
 sim_coupled = @time Speedy.initialize!(primitive_wet_coupled)
```
maximilian-gelbrecht:
```suggestion
sim_coupled = @time Speedy.initialize!(primitive_wet_coupled; time = initial_date)
```

maximilian-gelbrecht: As long as you start Jan 1 this won't make a difference, but better be save and add it here.

src/diagnostics/surface_fluxes.jl
```diff
@@ -0,0 +1,18 @@
+"""
+    $TYPEDSIGNATURES
+
+Return a `KernelFunctionOperation` that computes the skin temperature residual over all grid cells.
```
maximilian-gelbrecht, line 4: that's done because it's a lazy operation, right?

src/input_output/assets.jl
```diff
@@ -2,6 +2,8 @@
 abstract type FileFormat end
 struct NetCDF <: FileFormat end
 
+@enum AssetGrid native N72
```
maximilian-gelbrecht, line 5: I've never seen `@enum` before. Is it new or just fringe? 

I looked up what it does and it fits here, but just curious.

src/processes/atmosphere/prescribed_atmosphere.jl
```diff
@@ -81,7 +81,7 @@ end
 function PrescribedAtmosphere(
         ::Type{NF};
         altitude::NF = NF(10), # Default to 10 m
-        min_windspeed::NF = NF(0.01), # Default to 0.01 m/s
+        min_windspeed::NF = NF(0.01), # Default to 0.1 m/s
```
maximilian-gelbrecht: the changed comment contradicts the argument.

src/processes/surface/evapotranspiration/canopy_evapotranspiration.jl
```diff
@@ -47,18 +54,23 @@ Compute the transpiration vapor conductance [m/s] from aerodynamic resistance `r
 conductance `g_stm`. The transpiration flux is this conductance times the humidity gradient.
 """
 @inline function transpiration_conductance(::PALADYNCanopyEvapotranspiration{NF}, rₐ, g_stm) where {NF}
-    rₐ_can = 1 / max(g_stm, sqrt(eps(NF))) # clip stomatal conductance
-    return 1 / (rₐ + rₐ_can)
+    # From the general formula for conductances in series:
+    # g_total = g_1 * g_2 / (g_1 + g_2)
+    # substituting g_1 = 1 / r_1
+    g_trp = g_stm / (1 + g_stm * rₐ)
```
maximilian-gelbrecht, line 60: Seems like a quite big change in how it's computed. what caused it?

src/processes/surface/evapotranspiration/ground_resistance_factor.jl
```diff
@@ -46,19 +46,21 @@ ground_evaporation_resistance_factor(i, j, grid, fields, ::SoilMoistureResistanc
...
-    θres = props.residual * soil.saturation
+    por = porosity(comp)
+    θres = props.residual_saturation * por
     return ground_evaporation_resistance_factor(res, θw, θfc, θres)
 end
 
 @inline function ground_evaporation_resistance_factor(::SoilMoistureResistanceFactor{NF}, θw, θfc, θres) where {NF}
     if θw < θfc
-        β = (1 - cos(π * (θw - θres) / (θfc - θres)))^2 / 4
+        fc_sat = ifelse(θfc > θres, (θw - θres) / (θfc - θres), zero(NF)) # guard against corner case where θfc == θres
```
maximilian-gelbrecht: I am being very pedantic here but isn't it an edge case and not a corner case?

test/surface/skin_temperature.jl
maximilian-gelbrecht, line 1: I can't follow this in detail, what's the purpose of the rewrite here?

docs/dev/2026-07/2026-07-08-PLAN_reactant.md
bgroenks96: I guess this is a problem with having these files in the repository. They also get touched by every find/replace. Not sure if VSCode has a way to block certain folders from the search function...?

On the one hand, I do see your point that it's a bit strange and risks altering the record, On the other hand, it does at least keep the old plan documents a bit more up-to-date so they can remain at least somewhat relevant.

I would also be open to removing the plan documents from the repository if we can come up with some other place to store them so that they are transparent and available. We could maybe use the GitHub wiki page for this, but it would require an additional manual step. What do you think?

src/diagnostics/surface_fluxes.jl
bgroenks96: Yes, but also `KernelFunctionOperation` is just the canonical Oceananigans pattern for constructing diagnostics from kernel functions that do not have associated `Field`s.

src/input_output/assets.jl
bgroenks96: It's not new, I think it dates back to at least Julia 1.0, maybe earlier.

I never really understood why it isn't used more in Julia... I guess because people tend to prefer using abstract types that are extensible. But sometimes you explicitly do not want things to be extended. Enums are the right pattern in such cases.

bgroenks96: It's also a bit of a pet peeve of mine that people abuse `Symbol`s all the time as effectively enums, e.g. passing arguments like `arg = :value` where `arg` has a fixed number of possible options. I think this is an anti-pattern that should always be replaced by an enum.

src/processes/atmosphere/prescribed_atmosphere.jl
bgroenks96: :man_facepalming: debugging artifact...

src/processes/surface/evapotranspiration/canopy_evapotranspiration.jl
bgroenks96: It's actually mathematically equivalent. The change is only numerical. Before we added the resistances in series, i.e
```math
g_\text{trp} = \frac{1}{r_a + r_\text{stm}}
```
with $r_\text{stm} = \frac{1}{g_\text{stm}}$. For numerical reasons, I naively clipped $g_\text{stm}$ to prevent division by zero but didn't realize that this actually breaks the physics, because $g_\text{stm} = 0$ actually *should* result in infinite resistance to stop transpiration entirely.

This change just rewrites the combination of resistances in terms of conductances to avoid the division by zero. Per the comment,
```math
g_\text{total} = \frac{g_1 g_2}{g_1 + g_2} = \frac{1}{\frac{1}{g_1} + \frac{1}{g_2}}
```

Let $g_1 = g_a = \frac{1}{r_a}$, $g_2 = g_\text{stm}$, and $g_\text{total} = g_\text{trp}$. Substituting in the expression for $g_a$, we get
```math
g_\text{trp} = \frac{g_\text{stm}}{r_a (g_a + g_\text{stm})} = \frac{g_\text{stm}}{\frac{r_a}{r_a} + g_\text{stm} r_a} = \frac{g_\text{stm}}{1 + g_\text{stm} r_a}
```

So nothing changed other than removing the incorrect clipping and rewriting the equation to multiply by zero instead of divide. We probably could have kept the original formulation without the clipipng since `1 / Inf == 0` but I seem to recall having issues with Enzyme gradients in these cases, which was the reason for `safediv`.

src/processes/surface/evapotranspiration/ground_resistance_factor.jl
bgroenks96: I wasn't aware that there is a distinction between those two terms, but apparently yes, it's an edge case.

test/surface/skin_temperature.jl
bgroenks96: I wouldn't call it a rewrite. More of an expansion to cover more cases. It was motivated by the need to debug convergence failures in the SEB.

src/processes/surface/evapotranspiration/canopy_evapotranspiration.jl
bgroenks96: @olivierbonte in case you're interested

docs/dev/2026-07/2026-07-08-PLAN_reactant.md
maximilian-gelbrecht: Using the GitHub wiki for that sounds like a good idea, but we can set this up at another time. 

As far as I can see it the wiki is handled as a separate git repository. So it'll probably possible to set this up but it'll need a bit intricate setup than what we have right now.

src/processes/surface/evapotranspiration/canopy_evapotranspiration.jl
maximilian-gelbrecht: Thanks, good to know!

maximilian-gelbrecht (approved):

src/processes/atmosphere/prescribed_atmosphere.jl
bgroenks96: Fixed in [5e96b9b](https://github.com/NumericalEarth/Terrarium.jl/pull/170/commits/5e96b9b8f996e28d721cf7bdf7d7388188653455)

src/processes/surface/evapotranspiration/ground_resistance_factor.jl
bgroenks96: Fixed in [5e96b9b](https://github.com/NumericalEarth/Terrarium.jl/pull/170/commits/5e96b9b8f996e28d721cf7bdf7d7388188653455)

src/processes/surface/evapotranspiration/canopy_evapotranspiration.jl
bgroenks96: Also, since you asked "what caused it", the actual answer to that is that there was a bunch of transpiration happening over the Australian desert, and I figured that probably shouldn't happen ;)

PR #171, opened 2026-08-16
Reduce compile latency of Terrarium model initialization

### LINKED ISSUE #66: Issues with bloated model state and long compile times

bgroenks96:
Reducing compilation time is no easy feat when working with complex, nested types (with lots of `NamedTuple`s), but it has long been clear that there are some aspects of Terrarium's design that could be modified to mitigate it. We take those steps here. This effort is, in part, motivated by the somewhat painful debugging process of #170 thanks to the endemically long compile times of `initialize` and the first `timestep!`. This is further exacerbated by the fact that each call to `initialize` *without any change in arguments* still often triggers partial recompilation. See also #66.

This PR consists of two major changes:

1. **Refactoring of state variable initialization**. The `Variables` and `StateVariables` containers now use `OrderedDict`s to collect and merge the model variables and iteratively allocate `Field`s for each one. This greatly reduces type complexity for the compiler; we got a nearly 40% reduction in compile time from this change alone (~156 seconds --> ~93 seconds for `LandModel`).
2. **New top-level method `compute_boundary_conditions!`** that is intended to make the previously implicit BC operations (`fill_halo_regions!` and `compute_z_bcs!`) explicit; i.e. we only invoke them where needed rather than on the full `state`. This adds a bit of boilerplate code but results in another ~50% reduction in TTFX for `initialize` on `LandModel`, now down to ~47 seconds.

Part 2 is admittedly a somewhat significant design change, but is not without other motivations; `compute_boundary_conditions!` also acts as a more natural home for boundary condition physics like the SEB and snow/soil interface fluxes. This also partially addresses some concerns previously raised by @olivierbonte and @pierluigividale regarding the SEB solver being somewhat awkwardly shoehorned into `compute_auxiliary!`.

## Results

| Function | `main` (baseline) | Part I | Part II | Δ total | Δ allocs |
| --- | --- | --- | --- | --- | --- |
| `initialize(::LandModel)` | 157.6 s (42.15 M, 2.84 GiB, 104.3% compile) | 96.2 s (41.29 M, 2.79 GiB, 107.4% compile) | **47.7 s** (40.36 M, 2.73 GiB, 114.0% compile) | **-70%** | -4% |
| `timestep!(::ModelIntegrator)` | 109.7 s (15.00 M, 1.01 GiB, 101.1% compile) | 115.2 s (15.14 M, 1.02 GiB, 101.1% compile) | **74.7 s** (15.28 M, 1024 MiB, 101.7% compile) | **-32%** | +2% |

All compiler benchmarks are run with a full `LandModel` (snow + soil + surface energy/hydrology + vegetation carbon). These are simple `@time` outputs so they should not be regarded as robust estimates.

## TODO
- [x] Fix remaining broken tests

Resolves #66 (for now)


src/processes/snow/snow_single_layer.jl
```diff
@@ -93,7 +93,6 @@ variables(snow::SingleLayerSnow) = (
     input(:surface_heat_flux, XY(); units = u"W/m^2", desc = "Net heat flux at the snow surface (positive upward)"),
     input(:basal_heat_flux, XY(); units = u"W/m^2", desc = "Conductive heat flux at the snow base (positive upward, soil → snow)"),
     input(:sublimation, XY(); units = u"m/s", desc = "Sublimation/evaporation rate from the snow surface (SWE)"),
-    variables(snow.closure)...,
```
maximilian-gelbrecht, line 96: the closure can't have any variables?

src/models/coupled/land_model.jl
maximilian-gelbrecht, line 1: are these changes here related to the compile time?

src/processes/soil/energy/soil_energy.jl
```diff
@@ -70,6 +70,13 @@ end
 """ $TYPEDSIGNATURES """
 compute_auxiliary!(state, grid, energy::SoilThermodynamics, soil::AbstractSoil, args...) = nothing
 
+""" $TYPEDSIGNATURES """
+function compute_boundary_conditions!(state, grid, ::SoilThermodynamics)
```
maximilian-gelbrecht, line 74: how is this and the other `compute_boundary_conditions!` related to compile latency?

src/processes/surface/surface_energy_balance.jl
```diff
@@ -47,16 +47,12 @@ variables(seb::SurfaceEnergyBalance) = tuplejoin(
...
-        constants::PhysicalConstants,
-        atmos::AbstractAtmosphere,
-        hydrology::Optional{AbstractSurfaceHydrology} = nothing,
         vegetation::Optional{AbstractVegetation} = nothing,
         snow::Optional{AbstractSnow} = nothing,
         args...
     )
     # diagnose the (optionally snow-aware) albedo, then solve the surface energy balance
     compute_auxiliary!(state, grid, seb.albedo, vegetation, snow)
-    solve_surface_energy_balance!(state, grid, seb, constants, atmos, hydrology, snow)
```
maximilian-gelbrecht, line 59: why is this not needed anymore here?

src/timesteppers/abstract_timestepper.jl
```diff
@@ -148,10 +148,6 @@ function explicit_step!(state, grid::AbstractLandGrid, timestepper::AbstractTime
     # step only this namespace's prognostic variables that are also selected in `names`
     fastiterate(prognostic_names(state)) do name
         if name ∈ names
-            # apply flux BCs, if present
-            compute_z_bcs!(state.tendencies[name], state.prognostic[name], grid, state)
```
maximilian-gelbrecht, line 152: why is this not needed anymore?

maximilian-gelbrecht (commented):
Half OT: At a somewhat disappointing talk at JuliaCon about (pre-)compilation the top advice was also "just ask an LLM to reduce it for your package". This wasn't very satisfying.

src/processes/snow/snow_single_layer.jl
bgroenks96: No, it's just redundant because the `Variables` constructor already collects them directly. We could get rid of that functionality and then just always put the closures here in `variables` per-process like all of the other parameterizations. That would be more transparent, I guess.

src/models/coupled/land_model.jl
bgroenks96: The changes to the SEB are not related to compile time but were a natural consequence of having `compute_boundary_conditions!` available.

src/processes/soil/energy/soil_energy.jl
bgroenks96: As discussed in the PR description, the automatic `fill_halo_regions!` and `compute_z_bcs!` on the full `state` (via `fastmap`) had an outsized impact on compile time. It's not entirely clear to me why, but I guess it's because of a lot of downstream Oceananigans code getting compiled for no reason. It was also anyway wasteful in runtime as well, and the added boilerplate seems OK to me.

src/timesteppers/abstract_timestepper.jl
bgroenks96: Because it's done explicitly by the processes...

src/processes/surface/surface_energy_balance.jl
bgroenks96: Moved to `compute_boundary_conditions!`

src/models/soil/soil_model.jl
```diff
@@ -44,6 +44,11 @@ function compute_auxiliary!(state, model::SoilModel)
     return nothing
 end
 
+function compute_boundary_conditions!(state, model::SoilModel)
+    compute_boundary_conditions!(state, model.grid, model.soil)
```
maximilian-gelbrecht, line 48: Do we expect every model to implement this or is there a fallback? 

Are the expected signatures documented somewhere? It should be probably added to the docs as well.

src/processes/soil/energy/soil_energy.jl
maximilian-gelbrecht: But don't all still all fields need to call `fill_halo_regions!` in the end? 

I understand the intention, but I am a bit confused this makes a difference.

src/models/soil/soil_model.jl
maximilian-gelbrecht: Ah okay, found it below in `abstract_model.jl`, but should probably be still added somewhere to the docs.

src/abstract_variables.jl
```diff
@@ -284,58 +291,82 @@ with_scope(path::VarPath, var::AbstractVariable) =
...
+    for var in values(prognostic_vars)
+        tendency_vars[varname(var)] = var.tendency
+    end
+    # check for illegal duplicates across variable groups
+    check_duplicates(
+        values(prognostic_vars)...,
+        values(auxiliary_vars)...,
+        values(input_vars)...,
+        values(namespaces)...
+    )
```
maximilian-gelbrecht, line 358: I can see that this does the same as before, I assume it's done differently to reduce compile time. 

Is there something we can learn from this what to avoid in the future?

src/abstract_variables.jl
```diff
@@ -349,18 +380,15 @@ function check_duplicates(vars::Union{AbstractVariable, Namespace}...)
...
+Type-stable equivalent of [`deduplicate`](@ref) for tuples of `AbstractVariable`s and `Namespace`s.
 """
-function merge_namespaces(namespaces::Tuple{Vararg{Namespace}})
-    names = unique(map(varname, namespaces))
-    merged = map(names) do name
-        group = filter(ns -> varname(ns) == name, namespaces)
-        length(group) == 1 ? group[1] : Namespace(name, tuplejoin(map(ns -> ns.vars, group)...))
-    end
-    return Tuple(merged)
+@generated function deduplicate_vars(vars::Tuple{Vararg{Union{AbstractVariable, Namespace}}})
```
maximilian-gelbrecht, line 387: `deduplicate` isn't used above anymore, is it still used somewhere else?

src/processes/soil/energy/soil_energy.jl
bgroenks96: No. `fill_halo_regions!` is only necessary when the `Field` actually has boundary conditions that are used by some spatial operators. Doing it for all `Field`s in the model (actually just all prognostic and closure `Field`s, auxiliaries were already removed some time ago) was purely a user convenience feature, not a requirement for correctness.

src/abstract_variables.jl
bgroenks96: I guess just like in Speedy, use `Dict`-like types instead of `NamedTuple`s in all performance non-critical (and non-kernel) code? Unless the `NamedTuple`s have a known, bounded size and type complexity, in which case it shouldn't matter.

src/abstract_variables.jl
bgroenks96: I don't think so. Do you want to remove it?

I understand the motivation to want to remove dead code, but I also am sometimes hesitant to rely on future me remembering to look through the git history when I eventually need the functionality again.

maximilian-gelbrecht: I think, personally, I would remove it, but if you want to keep it I also don't mind.

maximilian-gelbrecht (approved):

src/abstract_variables.jl
bgroenks96: I actually forgot how minimal that function was. Removed in [9632e7e](https://github.com/NumericalEarth/Terrarium.jl/pull/171/commits/9632e7e33bc05632b650f88e0b978238baa89838).

src/models/soil/soil_model.jl
bgroenks96: Done in [8c17416](https://github.com/NumericalEarth/Terrarium.jl/pull/171/commits/8c174160917615064ac2aa9eba17e0061708e422). I also added some notes on `timestepper` being part of `AbstractModel`s, ~~as well as a `get_timestepper` method which was previously removed~~ (nevermind, I forgot that it's in `abstract_timestepper.jl`.

PR #173, opened 2026-08-21
Fix bugs in coupling of evapotranspiration and soil hydrology

### LINKED ISSUE #172: ET kinematic humidity flux is treated as liquid-water-depth flux in soil hydrology
## Summary

The evapotranspiration forcing applied to soil hydrology appears to treat a kinematic specific-humidity flux as a liquid-water-depth flux.

Although both quantities can be written dimensionally as `m/s`, they differ by the density ratio $\rho_{\rm air}/\rho_{\rm water}$. This means evaporation or condensation is applied to soil water roughly 800 times too much.

I encountered this using Terrarium v0.1.2, but the same dimensional inconsistency remains on current `main` at commit [`[97dd686](https://github.com/NumericalEarth/Terrarium.jl/commit/97dd686676afa726bb6917beb36dc52e106d93da)`](https://github.com/NumericalEarth/Terrarium.jl/commit/97dd686676afa726bb6917beb36dc52e106d93da).

### Kinematic humidity flux

[`[bare_ground_evaporation.jl:30–43](https://github.com/NumericalEarth/Terrarium.jl/blob/97dd686676afa726bb6917beb36dc52e106d93da/src/processes/surface/evapotranspiration/bare_ground_evaporation.jl#L30-L43)`](https://github.com/NumericalEarth/Terrarium.jl/blob/97dd686676afa726bb6917beb36dc52e106d93da/src/processes/surface/evapotranspiration/bare_ground_evaporation.jl#L30-L43) computes the surface humidity flux from a vapour conductance $g$ and a specific-humidity difference $\Delta q$:

```julia
Evaluate the bare-ground surface humidity flux [m/s] from the current skin temperature and
evaporation conductance in `fields`.
"""
@propagate_inbounds function compute_surface_humidity_flux(
        i, j, grid, fields,
        ::BareGroundEvaporation,
        constants::PhysicalConstants,
        atmos::AbstractAtmosphere,
        args...
    )
    Ts = fields.skin_temperature[i, j]
    g = fields.ground_evaporation_conductance[i, j]
    Δq = compute_specific_humidity_difference(i, j, grid, fields, atmos, constants, Ts)
    return g * Δq
end
```

Units are explicitly stated here [`[evapotranspiration_base.jl:17–25](https://github.com/NumericalEarth/Terrarium.jl/blob/97dd686676afa726bb6917beb36dc52e106d93da/src/processes/surface/evapotranspiration/evapotranspiration_base.jl#L17-L25)`](https://github.com/NumericalEarth/Terrarium.jl/blob/97dd686676afa726bb6917beb36dc52e106d93da/src/processes/surface/evapotranspiration/evapotranspiration_base.jl#L17-L25):

```julia
"""
    $TYPEDSIGNATURES

Compute an evapotranspiration flux (m/s, positive upwards) as the product of a vapor
conductance `g` (m/s) and a specific humidity difference `Δq` (kg/kg). All evapotranspiration components
— soil/canopy evaporation and transpiration — share the functional form ``E = Δq · g``,
differing only in which conductance and humidity difference are supplied.
"""
@inline compute_evaporation_flux(::AbstractEvapotranspiration, Δq, g) = Δq * g
```

But because $\Delta q$ is specific humidity in kg vapour per kg air, $g\Delta q$ is the kinematic tracer flux

$$
Q_h = g\Delta q,
$$

with units

$$
\frac{\mathrm{kg_{vapour}}}{\mathrm{kg_{air}}},\mathrm{m s^{-1}},
$$

But this is not actually metres of liquid water per second.

### Latent-heat calculation 

[`[turbulent_fluxes.jl:44–52](https://github.com/NumericalEarth/Terrarium.jl/blob/97dd686676afa726bb6917beb36dc52e106d93da/src/processes/surface/turbulent_fluxes.jl#L44-L52)`](https://github.com/NumericalEarth/Terrarium.jl/blob/97dd686676afa726bb6917beb36dc52e106d93da/src/processes/surface/turbulent_fluxes.jl#L44-L52) converts this humidity flux to latent heat by multiplying by air density:

```julia
"""
    $TYPEDSIGNATURES

Compute the latent heat flux as a function of the humidity flux `Q_h` [m/s], the density `ρₐ` [kg/m³] of air,
and the specific latent heat of vaporization or sublimation `L` [J/kg].
"""
function compute_latent_heat_flux(::DiagnosedTurbulentFluxes, Q_h, ρₐ, L)
    Hₗ = L * ρₐ * Q_h
    return Hₗ
end
```

The full calculation uses this expression at [`[turbulent_fluxes.jl:184–212](https://github.com/NumericalEarth/Terrarium.jl/blob/97dd686676afa726bb6917beb36dc52e106d93da/src/processes/surface/turbulent_fluxes.jl#L184-L212)`](https://github.com/NumericalEarth/Terrarium.jl/blob/97dd686676afa726bb6917beb36dc52e106d93da/src/processes/surface/turbulent_fluxes.jl#L184-L212):

```julia
Q_h = if isnothing(evtr)
    # Direct calculation of evaporative flux without ET coupling
    Tₛ = skin_temperature(i, j, grid, fields, skinT)
    rₐ = aerodynamic_resistance(i, j, grid, fields, atmos)
    Δq = compute_specific_humidity_difference(i, j, grid, fields, atmos, constants, Tₛ)
    Δq / rₐ
else
    # Compute humidity flux using the given ET scheme
    compute_surface_humidity_flux(i, j, grid, fields, evtr, constants, atmos)
end

# ...

ρₐ = Thermodynamics.air_density(constants.thermodynamics, celsius_to_kelvin(constants.thermodynamics, Tₐ), pres, q_air)
L_lv = constants.thermodynamics.latent_heat_vaporization

# ...

Hₗ_ground = compute_latent_heat_flux(tur, Q_h, ρₐ, L_lv)
```

Therefore Terrarium interprets $Q_h$ as a kinematic atmospheric humidity flux:

$$
H_L = L_v\rho_{\rm air}Q_h.
$$

### Soil hydrology instead treats the same value as liquid-water depth

The ET forcing in [`[evapotranspiration_base.jl:29–39](https://github.com/NumericalEarth/Terrarium.jl/blob/97dd686676afa726bb6917beb36dc52e106d93da/src/processes/surface/evapotranspiration/evapotranspiration_base.jl#L29-L39)`](https://github.com/NumericalEarth/Terrarium.jl/blob/97dd686676afa726bb6917beb36dc52e106d93da/src/processes/surface/evapotranspiration/evapotranspiration_base.jl#L29-L39) applies the raw (Q_h) directly:

```julia
"""
    forcing(i, j, k, grid, clock, fields, evapotranspiration::AbstractEvapotranspiration, ::AbstractSoilHydrology)

Compute and return the evapotranspiration forcing for soil moisture at the given indices `i, j, k`.
The ET forcing is just the `surface_humidity_flux` rescaled by the thickness of layer `k`.
"""
@inline function forcing(i, j, k, grid, clock, fields, evapotranspiration::AbstractEvapotranspiration, ::AbstractSoilHydrology)
    let Δz = Δzᵃᵃᶜ(i, j, k, grid)
        Qh = surface_humidity_flux(i, j, grid, fields, evapotranspiration)
        ∂θ∂t = -Qh / Δz # rescale by layer thickness to get water content flux
        return ∂θ∂t * (k == grid.Nz)
    end
end
```

This (I think) assumes that

$$
Q_h = E_{\rm water}
$$

is a liquid-water-depth flux in m water s^-1. But that's not consistent with the latent-heat calculation above.

The soil saturation tendency is subsequently divided by porosity in [`[soil_hydrology.jl:207–220](https://github.com/NumericalEarth/Terrarium.jl/blob/97dd686676afa726bb6917beb36dc52e106d93da/src/processes/soil/hydrology/soil_hydrology.jl#L207-L220)`](https://github.com/NumericalEarth/Terrarium.jl/blob/97dd686676afa726bb6917beb36dc52e106d93da/src/processes/soil/hydrology/soil_hydrology.jl#L207-L220):

```julia
@propagate_inbounds function compute_saturation_tendency!(
        saturation_water_ice_tendency, i, j, k, grid, clock, fields,
        hydrology::SoilHydrology,
        strat::AbstractStratigraphy,
        bgc::AbstractSoilBiogeochemistry,
        constants::PhysicalConstants,
        evtr::Optional{AbstractEvapotranspiration}
    )
    # Compute volumetic water content tendency
    ∂θ∂t = compute_volumetric_water_content_tendency(i, j, k, grid, clock, fields, hydrology, constants, evtr)
    # Get porosity
    por = porosity(i, j, k, grid, fields, strat, bgc)
    # Rescale by porosity to get saturation tendency
    saturation_water_ice_tendency[i, j, k] += ∂θ∂t / por
```

## Correct conversion?

The vapour mass flux should be

$$
F_q = \rho_{\rm air}Q_h
\qquad [\mathrm{kg~m^{-2}~s^{-1}}],
$$

and the equivalent liquid-water-depth flux should be

$$
E_{\rm water}
= \frac{F_q}{\rho_{\rm water}}
= \frac{\rho_{\rm air}}{\rho_{\rm water}}Q_h
= \frac{H_L}{L_v\rho_{\rm water}}
\qquad [\mathrm{m~s^{-1}}].
$$

The hydrological forcing should therefore be

$$
\frac{\partial\theta}{\partial t}
= -\frac{E_{\rm water}}{\Delta z},
$$

rather than $-Q_h/\Delta z$.

Either

1. If $Q_h$ is the kinematic humidity flux, the latent-heat calculation is correct and the hydrological forcing needs the factor $\rho_{\rm air}/\rho_{\rm water}$.
2. If $Q_h$ is already liquid-water depth, the hydrological forcing is correct but latent heat would have to be $L_v\rho_{\rm water}Q_h$, not $L_v\rho_{\rm air}Q_h$.

I think (1) is correct

## Quick example

In a one-column Terrarium v0.1.2 test with matched soil, skin and air temperatures of 15°C, the diagnosed values were

```text
surface humidity flux Qh = -1.13934e-6 m s-1
latent heat flux HL      = -3.13089 W m-2
top-layer thickness      = 0.05 m
porosity                  = 0.49
```

Terrarium v0.1.2 used

```text
Lv    = 2.257e6 J kg-1
rho_w = 1000 kg m-3
```

The latent heat implies

$$
F_q = \frac{H_L}{L_v}
= -1.38719\times10^{-6}\ \mathrm{kg~m^{-2}~s^{-1}},
$$

and therefore

$$
E_{\rm water}
= \frac{H_L}{L_v\rho_{\rm water}}
= -1.38719\times10^{-9}\ \mathrm{m~s^{-1}}.
$$

The condensation rate is consequently about 0.120 mm/d.

Treating $Q_h$ as liquid-water depth instead gives approximately 98.4 mm/d of condensation.

The old top-layer saturation tendency is reproduced exactly by the current formulation:

$$
\frac{|Q_h|}{0.05\times0.49}
= 4.65037\times10^{-5}\ \mathrm{s^{-1}}.
$$

Using the liquid-water conversion gives

$$
\frac{|H_L|}
{L_v\rho_{\rm water}(0.05)(0.49)}
=5.6620\times10^{-8}\ \mathrm{s^{-1}}.
$$

Adding the independently diagnosed pure-Richards tendency,

$$
2.81464\times10^{-9}\ \mathrm{s^{-1}},
$$

produces

$$
5.9435\times10^{-8}\ \mathrm{s^{-1}},
$$

which matches the corrected one-column result. This is a reduction of approximately 780x in the total top-layer tendency.

bgroenks96:
This PR addresses #172 as well as some other related issues:

1. ~~The soil hydrology + ET coupling is now refactored to use a separate field `evaporation_soil_water` which is treated as an input by soil hydrology and auxiliary variable by ET. The auxiliary variable is currently declared by `LandModel` in `interface_variables`. This separate term is rescaled by the ratio of air to water density as described in #172. The allocation of a new `Field` is not ideal (the original motivation of using `forcing` was exactly to avoid this) but necessary for now to avoid making soil hydrology dependent on atmospheric state (air density).~~
All evaporative fluxes (e.g .`evaporation_ground`, `evaporation_canopy`, and `transpiration`) are now computed and stored as liquid water mass fluxes, i.e. $E = Q_h \cdot \frac{\rho_a}{\rho_w}$. This avoids the aforementioned need to define a new `Field` since all downstream hydrology fluxes (including evaporation of canopy water which was also previously using the humidity flux) can directly consume $E$ without needing to rescale by the density ratio, which requires calculation of $\rho_a$. The surface energy balance still uses the raw kinematic humidity flux $Q_h$, so this shouldn't affect the coupling with SpeedyWeather.
2. The soil hydrology ET forcing previously used the full `surface_humidity_flux`; this was actually wrong because this includes evaporation from the canopy which draws from a separate water pool. This is now corrected, and the `forcing` implementation for `PALADYNCanopyEvapotranspiration` now includes only the ground evaporation and transpiration fluxes. This bug did not affect `BareGroundEvaporation`.
3. For `BareGroundEvaporation`, snow-covered fraction scaling was wrongly applied twice, both in `compute_evapotranspiration_fluxes!` as well as in `compute_latent_heat_flux!`. This is also now fixed; ~~scaling is done once at the source (the ET code) and `compute_latent_heat_flux!` assumes the humidity flux and sublimation flux to already be scaled.~~ Evaporative fluxes are area-weighted in `compute_evapotranspiration_fluxes!` but the raw kinematic humidity and sublimation fluxes are now left unscaled. Thus, `compute_latent_heat_flux!` applies the area-weighting once in the final latent heat flux calculation.

## Additional changes
- We now adopt a more consistent notational convention to hopefully reduce the chance of this mistake occurring in the future: evaporation (liquid water mass) fluxes are denoted as $E$ (e.g. `E_gnd`, `E_trp`, and `E_can`) while kinematic humidity fluxes are denoted as $Q_h$ (e.g. `Qh_gnd`, `Qh_trp`, and `Qh_can`).
- Sublimation is no longer computed with area-weighting in `compute_snow_sublimation_flux`; this applied later in the final computation of `sublimation`. In a future PR, we should also consider moving `sublimation` into the ET code to keep all humidity flux calculations in one place.

Resolves #172

### DIFF
--- a/src/processes/snow/mass/snow_mass.jl
+++ b/src/processes/snow/mass/snow_mass.jl
@@ -3,12 +3,11 @@
 """
     $TYPEDSIGNATURES
 
-Snow-surface sublimation rate [m/s SWE] at grid cell `i, j`, area-weighted by the snow-covered fraction
-`f_snow`. The snow surface is treated as saturated: a bulk-aerodynamic vapor flux `Δq/rₐ` evaluated at
-the skin temperature, with `Δq` taken over ice for a sub-freezing surface (the saturation humidity
-already dispatches over ice for `T ≤ 0` — see [`saturation_specific_humidity_vapor`](@ref)). The
-water-vapor mass flux `ρₐ·Δq/rₐ` is converted to a snow-water-equivalent rate via `ρ_w`. Zero without
-snow (`snow === nothing`).
+Snow-surface sublimation rate [m/s SWE] at grid cell `i, j`. The snow surface is treated as saturated:
+a bulk-aerodynamic vapor flux `Δq/rₐ` evaluated at the skin temperature, with `Δq` taken over ice for
+a sub-freezing surface (the saturation humidity already dispatches over ice for `T ≤ 0` — see [`saturation_specific_humidity_vapor`](@ref)).
+The water-vapor mass flux `ρₐ·Δq/rₐ` is converted to a snow-water-equivalent rate via `ρ_w`. Zero without
+snow (i.e. when `snow === nothing`).
 """
 @propagate_inbounds compute_snow_sublimation_flux(i, j, grid, fields, ::Nothing, atmos, constants, skinT) = zero(eltype(grid))
 @propagate_inbounds function compute_snow_sublimation_flux(
@@ -18,7 +17,6 @@ snow (`snow === nothing`).
         constants::PhysicalConstants,
         skinT::AbstractSkinTemperature
     )
-    f = snow_cover_fraction(i, j, grid, fields, snow)
     Tₛ = skin_temperature(i, j, grid, fields, skinT)
     rₐ = aerodynamic_resistance(i, j, grid, fields, atmos)
     Δq = compute_specific_humidity_difference(i, j, grid, fields, atmos, constants, Tₛ) # over ice for Tₛ ≤ 0
@@ -26,7 +24,7 @@ snow (`snow === nothing`).
     ρ_w = constants.material.density_water
     # saturated vapor mass flux converted to a snow-water-equivalent rate, area-weighted by the snow-covered
     # fraction `f` to give the grid-cell-mean sublimation (W_snow and Ū_snow are grid-cell means)
-    E_subl = f * ρ_a * (Δq / rₐ) / ρ_w
+    E_subl = ρ_a * (Δq / rₐ) / ρ_w
     return E_subl
 end
 
--- a/src/processes/snow/mass/snow_mass.jl
+++ b/src/processes/snow/mass/snow_mass.jl
@@ -26,7 +26,8 @@ snow (`snow === nothing`).
     ρ_w = constants.material.density_water
     # saturated vapor mass flux converted to a snow-water-equivalent rate, area-weighted by the snow-covered
     # fraction `f` to give the grid-cell-mean sublimation (W_snow and Ū_snow are grid-cell means)
-    return f * ρ_a * (Δq / rₐ) / ρ_w
+    E_subl = f * ρ_a * (Δq / rₐ) / ρ_w
+    return E_subl
 end
 
 # Kernel functions
--- a/src/processes/snow/mass/snow_mass.jl
+++ b/src/processes/snow/mass/snow_mass.jl
@@ -22,14 +22,11 @@ snow (`snow === nothing`).
     Tₛ = skin_temperature(i, j, grid, fields, skinT)
     rₐ = aerodynamic_resistance(i, j, grid, fields, atmos)
     Δq = compute_specific_humidity_difference(i, j, grid, fields, atmos, constants, Tₛ) # over ice for Tₛ ≤ 0
-    Tₐ = air_temperature(i, j, grid, fields, atmos)
-    pres = air_pressure(i, j, grid, fields, atmos)
-    q_air = specific_humidity(i, j, grid, fields, atmos)
-    ρₐ = Thermodynamics.air_density(constants.thermodynamics, celsius_to_kelvin(constants.thermodynamics, Tₐ), pres, q_air)
+    ρ_a = air_density(i, j, grid, fields, atmos, constants)
     ρ_w = constants.material.density_water
     # saturated vapor mass flux converted to a snow-water-equivalent rate, area-weighted by the snow-covered
     # fraction `f` to give the grid-cell-mean sublimation (W_snow and Ū_snow are grid-cell means)
-    return f * ρₐ * (Δq / rₐ) / ρ_w
+    return f * ρ_a * (Δq / rₐ) / ρ_w
 end
 
 # Kernel functions
--- a/src/processes/soil/hydrology/soil_hydrology_rre.jl
+++ b/src/processes/soil/hydrology/soil_hydrology_rre.jl
@@ -109,7 +109,7 @@ is thus not the same as the saturation tendency.
     field_grid = get_field_grid(grid)
 
     # Compute divergence of water fluxes
-    # ∂θ∂t = ∇⋅K(θ)∇Ψ + S, where Ψ = ψₘ + ψₕ + ψz, and S represents all sources and sinks such as ET losses
+    # ∂θ∂t = ∇⋅K(θ)∇Ψ + S, where Ψ = ψₘ + ψₕ + ψz, and S is a forcing term for sources and sinks such as ET losses
     ∂θ∂t = (
         - ∂zᵃᵃᶜ(i, j, k, field_grid, darcy_flux, fields.pressure_head, fields.hydraulic_conductivity)  # Darcy flux
             + forcing(i, j, k, grid, clock, fields, evapotranspiration, hydrology, constants)          # ET source/sink
--- a/src/processes/soil/hydrology/soil_hydrology_rre.jl
+++ b/src/processes/soil/hydrology/soil_hydrology_rre.jl
@@ -22,6 +22,7 @@ variables(hydrology::SoilHydrology{NF, RichardsEq}) where {NF} = (
     prognostic(:surface_excess_water, XY(), units = u"m", desc = "Excess water at the soil surface in m³/m²"),
     auxiliary(:hydraulic_conductivity, XYZ(z = Face()), units = u"m/s", desc = "Hydraulic conductivity of soil volumes in m/s"),
     auxiliary(:water_table, XY(), units = u"m", desc = "Elevation of the water table in meters"),
+    input(:evaporation_soil_water, XY(), units = u"m/s", desc = "Evapotranspiration contribution to soil moisture flux"),
     input(:liquid_water_fraction, XYZ(), default = NF(1), bounds = UnitInterval, desc = "Fraction of unfrozen water in the pore space"),
 )
 
@@ -109,11 +110,11 @@ is thus not the same as the saturation tendency.
     field_grid = get_field_grid(grid)
 
     # Compute divergence of water fluxes
-    # ∂θ∂t = ∇⋅K(θ)∇Ψ + forcing, where Ψ = ψₘ + ψₕ + ψz, and "forcing" represents sources and sinks such as ET losses
+    # ∂θ∂t = ∇⋅K(θ)∇Ψ + S, where Ψ = ψₘ + ψₕ + ψz, and S represents all sources and sinks such as ET losses
     ∂θ∂t = (
-        - ∂zᵃᵃᶜ(i, j, k, field_grid, darcy_flux, fields.pressure_head, fields.hydraulic_conductivity)
-            + forcing(i, j, k, grid, clock, fields, evapotranspiration, hydrology, constants) # ET forcing
-            + forcing(i, j, k, grid, clock, fields, hydrology.vwc_forcing, hydrology) # generic user-defined forcing
+        - ∂zᵃᵃᶜ(i, j, k, field_grid, darcy_flux, fields.pressure_head, fields.hydraulic_conductivity)  # Darcy flux
+            + forcing(i, j, k, grid, fields, evapotranspiration, hydrology, constants)                 # ET source/sink
+            + forcing(i, j, k, grid, clock, fields, hydrology.vwc_forcing, hydrology)                  # generic user-defined forcing
     )
     return ∂θ∂t
 end
--- a/src/processes/soil/hydrology/soil_hydrology_rre.jl
+++ b/src/processes/soil/hydrology/soil_hydrology_rre.jl
@@ -22,7 +22,6 @@ variables(hydrology::SoilHydrology{NF, RichardsEq}) where {NF} = (
     prognostic(:surface_excess_water, XY(), units = u"m", desc = "Excess water at the soil surface in m³/m²"),
     auxiliary(:hydraulic_conductivity, XYZ(z = Face()), units = u"m/s", desc = "Hydraulic conductivity of soil volumes in m/s"),
     auxiliary(:water_table, XY(), units = u"m", desc = "Elevation of the water table in meters"),
-    input(:evaporation_soil_water, XY(), units = u"m/s", desc = "Evapotranspiration contribution to soil moisture flux"),
     input(:liquid_water_fraction, XYZ(), default = NF(1), bounds = UnitInterval, desc = "Fraction of unfrozen water in the pore space"),
 )
 
--- a/src/processes/soil/hydrology/soil_hydrology_rre.jl
+++ b/src/processes/soil/hydrology/soil_hydrology_rre.jl
@@ -112,7 +112,7 @@ is thus not the same as the saturation tendency.
     # ∂θ∂t = ∇⋅K(θ)∇Ψ + S, where Ψ = ψₘ + ψₕ + ψz, and S represents all sources and sinks such as ET losses
     ∂θ∂t = (
         - ∂zᵃᵃᶜ(i, j, k, field_grid, darcy_flux, fields.pressure_head, fields.hydraulic_conductivity)  # Darcy flux
-            + forcing(i, j, k, grid, fields, evapotranspiration, hydrology, constants)                 # ET source/sink
+            + forcing(i, j, k, grid, clock, fields, evapotranspiration, hydrology, constants)          # ET source/sink
             + forcing(i, j, k, grid, clock, fields, hydrology.vwc_forcing, hydrology)                  # generic user-defined forcing
     )
     return ∂θ∂t
--- a/src/processes/surface/evapotranspiration/bare_ground_evaporation.jl
+++ b/src/processes/surface/evapotranspiration/bare_ground_evaporation.jl
@@ -19,7 +19,7 @@ BareGroundEvaporation(
     ground_resistance::GR = SoilMoistureResistanceFactor(NF)
 ) where {NF, GR} = BareGroundEvaporation{NF, GR}(ground_resistance)
 
-@propagate_inbounds surface_humidity_flux(i, j, grid, fields, evaporation::BareGroundEvaporation, args...) = fields.evaporation_ground[i, j]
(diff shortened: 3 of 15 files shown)


src/processes/snow/mass/snow_mass.jl
```diff
@@ -22,14 +22,12 @@ snow (`snow === nothing`).
...
-    pres = air_pressure(i, j, grid, fields, atmos)
-    q_air = specific_humidity(i, j, grid, fields, atmos)
-    ρₐ = Thermodynamics.air_density(constants.thermodynamics, celsius_to_kelvin(constants.thermodynamics, Tₐ), pres, q_air)
+    ρ_a = air_density(i, j, grid, fields, atmos, constants)
     ρ_w = constants.material.density_water
     # saturated vapor mass flux converted to a snow-water-equivalent rate, area-weighted by the snow-covered
     # fraction `f` to give the grid-cell-mean sublimation (W_snow and Ū_snow are grid-cell means)
-    return f * ρₐ * (Δq / rₐ) / ρ_w
+    E_subl = f * ρ_a * (Δq / rₐ) / ρ_w
+    return E_subl
```
maximilian-gelbrecht, line 28: This is just a refactor so that it can be reused easier and no actual change, or am I overlooking something? Not disagreeing though.

maximilian-gelbrecht (approved):
To be honest, I don't understand this well without looking deeper into this, but it looks to me on the surface, so go ahead and merge it when you are ready.

src/processes/snow/mass/snow_mass.jl
bgroenks96: Just formatting, I'm trying to avoid mathematical expressions in `return` statements.

src/processes/surface/evapotranspiration/bare_ground_evaporation.jl
```diff
@@ -110,20 +110,29 @@ end
...
-    out.evaporation_ground[i, j, 1] = (one(f_snow) - f_snow) * compute_evaporation_flux(evaporation, Δq, g_gnd)
+    E_gnd = (NF(1) - f_snow) * compute_evaporation_flux(evaporation, Δq, g_gnd)
+    # Store relative contributions
+    out.evaporation_ground[i, j, 1] = E_gnd
+    if !isnothing(soil)
+        # Get air/water densities
+        ρ_a = air_density(i, j, grid, fields, atmos, constants)
+        ρ_w = constants.material.density_water
+        # Compute evaporation soil water
+        out.evaporation_soil_water[i, j, 1] = E_gnd * ρ_a / ρ_w
```
bgroenks96: @kieranmrhunt Please check

src/processes/snow/mass/snow_mass.jl
maximilian-gelbrecht: I actually meant the whole thing including the air density. But I basically already saw where it was reused, so never mind.

src/processes/surface/evapotranspiration/canopy_evapotranspiration.jl
```diff
@@ -222,18 +222,28 @@ for the given scheme `evapotranspiration` and process dependencies.
...
+    # Get air/water densities
+    ρ_a = air_density(i, j, grid, fields, atmos, constants)
+    ρ_w = constants.material.density_water
 
-    # Store fluxes in corresponding output Fields
-    out.evaporation_ground[i, j, 1] = E_gnd
-    out.transpiration[i, j, 1] = E_trp
-    out.evaporation_canopy[i, j, 1] = E_can
+    # Compute evaporation soil water
+    out.evaporation_soil_water[i, j, 1] = (E_gnd + E_trp) * ρ_a / ρ_w
```
bgroenks96: @kieranmrhunt Please check

src/processes/snow/mass/snow_mass.jl
bgroenks96: Oh, yes that was refactoring to avoid repeating the same code everywhere.

src/processes/surface/evapotranspiration/bare_ground_evaporation.jl
```diff
@@ -108,22 +106,37 @@ bare-ground `evaporation` scheme.
...
+    Qh_gnd = compute_surface_humidity_fluxes(i, j, grid, fields, evaporation, constants, atmos)
     # Ground evaporation occurs only over the snow-free fraction (1 − f_snow); the snow-covered fraction
     # sublimates instead (see `compute_snow_sublimation_flux`). No scaling without snow (f_snow = 0).
     f_snow = snow_cover_fraction(i, j, grid, fields, snow)
-    out.evaporation_ground[i, j, 1] = (one(f_snow) - f_snow) * compute_evaporation_flux(evaporation, Δq, g_gnd)
+    # Get air/water densities
+    ρ_a = air_density(i, j, grid, fields, atmos, constants)
+    ρ_w = constants.material.density_water
+    # Scale Qh_gnd by snow-free fraction and and scale by air-water density ratio to get evaporative flux E_gnd
+    E_gnd = (NF(1) - f_snow) * Qh_gnd * ρ_a / ρ_w
```
bgroenks96, line 138: @kieranmrhunt Please review

src/processes/surface/evapotranspiration/canopy_evapotranspiration.jl
```diff
@@ -222,18 +216,26 @@ for the given scheme `evapotranspiration` and process dependencies.
...
+
+    # Get air/water densities
+    ρ_a = air_density(i, j, grid, fields, atmos, constants)
+    ρ_w = constants.material.density_water
+    r = ρ_a / ρ_w # ratio of air to water density [-]
+
+    # Rescale by snow-covered fraction (if applicable) and convert to liquid water flux
+    f_snow = snow_cover_fraction(i, j, grid, fields, snow)
+    f_bare = NF(1) - f_snow
+    out.evaporation_ground[i, j, 1] = f_bare * Qh_gnd * r
```
bgroenks96, line 236: @kieranmrhunt Please review

src/processes/surface/evapotranspiration/canopy_evapotranspiration.jl
```diff
@@ -204,14 +201,11 @@ for the given scheme `evapotranspiration` and process dependencies.
...
 
-    # Compute and store the partitioned ET fluxes from the current skin temperature. The ground
-    # evaporation is driven by the ground temperature; the canopy evaporation and transpiration are
-    # skin-driven. Re-running this kernel after the SEB solve (see `LandModel`) refreshes the
-    # skin-driven fluxes so they are consistent with the converged skin temperature.
-    E_gnd = compute_evaporation_flux(evapotranspiration, Δqg, g_gnd)
-    E_trp = compute_evaporation_flux(evapotranspiration, Δqs, g_trp)
-    E_can = compute_evaporation_flux(evapotranspiration, Δqs, g_can)
-    return E_gnd, E_trp, E_can
+    # Compute the
```
maximilian-gelbrecht: orphaned comment

src/processes/soil/hydrology/soil_hydrology_rre.jl
```diff
@@ -109,11 +109,11 @@ is thus not the same as the saturation tendency.
     field_grid = get_field_grid(grid)
 
     # Compute divergence of water fluxes
-    # ∂θ∂t = ∇⋅K(θ)∇Ψ + forcing, where Ψ = ψₘ + ψₕ + ψz, and "forcing" represents sources and sinks such as ET losses
+    # ∂θ∂t = ∇⋅K(θ)∇Ψ + S, where Ψ = ψₘ + ψₕ + ψz, and S represents all sources and sinks such as ET losses
```
maximilian-gelbrecht: Again just being a bit pedantic here but why did you change the naming in the comments away from `forcing` but we still use `forcing` as the name of the function?

src/processes/surface/turbulent_fluxes.jl
```diff
@@ -192,24 +192,21 @@ snow-covered fraction sublimates from the snowpack (latent heat of sublimation,
         compute_surface_humidity_flux(i, j, grid, fields, evtr, constants, atmos)
     end
 
-    # Get atmospheric variables and constants
-    Tₐ = air_temperature(i, j, grid, fields, atmos)
-    pres = air_pressure(i, j, grid, fields, atmos)
-    q_air = specific_humidity(i, j, grid, fields, atmos)
+    # Get atmospheric variables and snow cover fraction
     # TODO: density should be evaluated at surface temperature for better accuracy
```
maximilian-gelbrecht, line 196: what's the problem with doing that?

maximilian-gelbrecht (approved):
just some very minor comments

src/processes/soil/hydrology/soil_hydrology_rre.jl
bgroenks96: I thought it was weird when seeing it again that the comment was mixing mathematical notation with prose. I can re-add forcing to the prose after it though.

src/processes/surface/turbulent_fluxes.jl
bgroenks96: I don't know, I think @olivierbonte originally added this comment, or I added it in response to one of his comments. Maybe he can clarify?

src/processes/soil/hydrology/soil_hydrology_rre.jl
bgroenks96: Fixed in [c3591dd](https://github.com/NumericalEarth/Terrarium.jl/pull/173/commits/c3591dd62d97823a47791e342cc92a44d36f1f17)

src/processes/surface/evapotranspiration/canopy_evapotranspiration.jl
bgroenks96: Fixed in [c3591dd](https://github.com/NumericalEarth/Terrarium.jl/pull/173/commits/c3591dd62d97823a47791e342cc92a44d36f1f17)

src/processes/surface/turbulent_fluxes.jl
bgroenks96: Resolving this for now since it's out-of-scope of this PR, but @olivierbonte if you could make an issue clarifying the what needs to be improved here, that would be great!

PR #176, opened 2026-08-25
Reactant Vegetation model (without interpolation) 

maximilian-gelbrecht:
This runs the prescribed vegetation model with Reactant. 

This doesn't yet use FieldTimeSeries input, their interpolation under Reactant is another thing we might have to tackle, potentially upstream in Oceananigans. A did look briefly into this already, and this also doesn't seem terribly hard though. 

This depends on https://github.com/CliMA/Oceananigans.jl/pull/5880, as you can see there are really no other further changes needed here.


maximilian-gelbrecht:
This still depends on the upstream feature branch https://github.com/CliMA/Oceananigans.jl/pull/5949 but hopefully this will be merged and released in the next days. 

With it now everything runs with Reactant including prescribed LAI and the default `LandModel` as well, with the one exception of the root solver. 

I'd like to just have a different default root solver for Reactant so that users wouldn't need to construct the land model so manual, but we only have the information of `NF` not of the architecture when constructing.

maximilian-gelbrecht:
Failing CI should be resolved by #197

Issue #178, opened 2026-08-25
Full integration with Oceananigans `AbstractGrid` interface with support for multi-domain vertical discretizations


bgroenks96:
Our current grid implementations, `ColumnGrid` and `ColumnRingGrid`, both rely on an Oceananigans `RectilinearGrid` where all grid points are folded into the X-dimension. The Y-dimension topology is set to `Flat` and thus ignored. This was a pragmatic solution that let us easily integrate with SpeedyWeather and RingGrids without having to worry about regridding.

However, the implementation of a suitable `ComponentExchanger` for NumericalEarth `EarthSystemModel`s (see NumericalEarth/NumericalEarth.jl#148) requires Terrarium grids to be proper `Oceananigans` `AbstractGrid` types and also support the use of spatially explicit grid types like `LatitudeLongitudeGrid` to integrate with `DataWrangling`.

In addition, we also need a solution to handle different grid domains (mentioned in #145). Our current scheme assumes all `XYZ` variables to be defined in the soil domain; however, the eventual addition of multi-layer snow schemes and multi-layer canopies will require the ability to define variables with differing vertical layers. I think we can safely restrict this to basically three domains: **Soil** (or "Ground"), **Snow**, and **Canopy**. Soil and snow could also be combined into a single domain with a combined `MutableVerticalDiscretization`. However, I think it's probably better to separate them because soil and snow physics differ quite a bit and we probably want to avoid branching in combined kernels.

This issue can be addressed in essentially three parts:
1. Refactoring of `ColumnGrid` (rename to `ColumnLandGrid`) and `ColumnRingGrid` (rename to `GlobalColumnLandGrid`) to implement `AbstractGrid` interface and addition of `VarDomain` interface with three subtypes: `Ground`, `Snow`, and `Canopy`.
2. Implementation of `LandGrid <: AbstractLandGrid` as a generic 3D grid that wraps any underlying Oceanangains grid (like `LatitudeLongitudeGrid`)
3. Refactoring of all current processes' `variables` implementations to explicitly declare the variable domains in addition to the existing variable dimensions.

## Additional changes
- Replace all existing Terrarium vertical discretizations in favor of the built-in ones from Oceananigans
- Replace `get_field_grid` with domain getters: e.g. `get_ground_grid`, `get_snow_grid`, `get_canopy_grid`

Happy to get feedback on naming and general design.

CC @glwagner and @xkykai

PR #179, opened 2026-08-26
Make `AbstractLandGrid` subtype `AbstractGrid`

### LINKED ISSUE #181: Refactor existing `AbstractLandGrid` types to subtype `Oceananigans.Grids.AbstractGrid`
This is phase 1 of #178

Add topology type parameters to `AbstractLandGrid` and forward interface methods to the underlying grid. No new types needed.

```julia
# Before
abstract type AbstractLandGrid{NF, Arch} end

# After
abstract type AbstractLandGrid{NF, TX, TY, TZ, Arch} <: Oceananigans.AbstractGrid{NF, TX, TY, TZ, Arch} end
```

Forwarded methods (delegate to `get_field_grid(grid)`): `size`, `halo_size`, `nodes`, `architecture`, `isrectilinear`.

bgroenks96:
This was actually easier than I thought it would be. This is the first step in addressing #178.

This PR also adds the implementation plan for #178.

Resolves #181

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -248,9 +248,7 @@ Follow [ColPrac](https://github.com/SciML/ColPrac). Feature branches, descriptiv
 
 ## Implementation plans
 
-All major feature additions, bug fixes, or refactoring that requires substantial changes to the existing code must be prefaced with an **implementation plan** that is **reviewed and signed off by a human**. These plan documents should be organized by date and stored in `docs/dev/YYYY-MM`.
-
-You are **ABSOLUTELY REQUIRED** to get the explicit approval of a human developer before executing the implementation. The human developer must type their name after **Approval** along with the revision number. Old plans prior the addition of this provision are not subject to this policy.
+All major feature additions, bug fixes, or refactoring that requires substantial changes to the existing code must be prefaced with an **implementation plan** that is **reviewed and signed off by a human**. These plan documents should be organized by date and stored in `docs/dev/YYYY-MM`. You are **ABSOLUTELY REQUIRED** to get the explicit approval of a human developer before executing the implementation.
 
 Each document should be prefaced by the following template:
 
@@ -262,8 +260,6 @@ Each document should be prefaced by the following template:
 Date of initial draft: YYYY-MM-dd
 
 Base revision: <SHA1 of HEAD when plan was drafted>
-
-Approval: PENDING
 
 ## Originating prompt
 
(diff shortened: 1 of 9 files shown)


AGENTS.md
```diff
@@ -248,7 +248,11 @@ Follow [ColPrac](https://github.com/SciML/ColPrac). Feature branches, descriptiv
 
 ## Implementation plans
 
-All major feature additions, bug fixes, or refactoring that requires substantial changes to the existing code must be prefaced with an **implementation plan** that is reviewed and signed off by a human. These plan documents should be organized by date and stored in `docs/dev/YYYY-MM`. Each document should be prefaced by the following template:
+All major feature additions, bug fixes, or refactoring that requires substantial changes to the existing code must be prefaced with an **implementation plan** that is **reviewed and signed off by a human**. These plan documents should be organized by date and stored in `docs/dev/YYYY-MM`.
```
bgroenks96: @maximilian-gelbrecht Are you OK with this change to the implementation plan rules?

maximilian-gelbrecht: I don't mind it directly, but why do you think it's needed? I never had the problem that it just started to implement stuff when I explicitly just asked for a plan.

I also switch between the different modes: manual / plan / auto. And would basically only put it on auto when I give it the go to execute a plan.

bgroenks96: I actually more intended it as a way to force myself to actually read the plan in detail.

bgroenks96: but I can remove it if you don't think it's necessary.

ext/TerrariumReactantExt/TerrariumReactantExt.jl
```diff
@@ -22,7 +22,7 @@ const RARCH = ReactantState
 @inline Terrarium.uses_reactant(::Terrarium.ReactantMarker) = true
 
 # Land grids that live on the device.
-const ReactantLandGrid{NF} = AbstractLandGrid{NF, <:RARCH}
+const ReactantLandGrid{NF, TX, TY, TZ} = AbstractLandGrid{NF, TX, TY, TZ, <:RARCH}
```
maximilian-gelbrecht, line 25: This was the culprit. With the new grid, this needed adjustment too

AGENTS.md
bgroenks96: Reverted in [c1042fb](https://github.com/NumericalEarth/Terrarium.jl/pull/179/commits/c1042fbc58fc7a066d57874fb3d3a2e8447c63c1)

maximilian-gelbrecht (approved):

PR #184, opened 2026-08-27
Rework handling of snow-ground interface in surface energy balance

### LINKED ISSUE #177: Investigate issues with snow melting too fast in coupled `LandModel` simulations
It seems that, when vegetation is included, snow doesn't accumulate properly in the northern latitudes during winter time:

https://github.com/user-attachments/assets/ac55ada9-8f1d-41e5-a2db-84a44c63d325

This is likely an issue with how the surface energy balance is interacting with the snow pack. One issue is that the current implementation uses the `ground_heat_flux` to force the *top* of the snowpack; this conceptually makes sense when a snowpack is present, since the snow serves as the interface at which the SEB should be solved.

However, it has a problem when considering fractional snow cover (as we currently do). Namely, in the absence of proper tiling, we use fractionally weighted inputs (e.g. ground temperature, thermal conductivity, etc.) to calculate a bulk ground heat flux for the whole grid cell which is then used to force the energy tendency of the snowpack over the snow-covered part. This likely results in unrealistic fluxes for the snow.

The ideal approach would be to use tiling, where we solve the SEB separately for both the snow-covered and snow-free fractions of the grid cell. However, another interim solution might be to return to a formulation where the `ground_heat_flux` is fixed to the actual ground surface and then partition the ground heat flux in the SEB:
```math
R_\text{net}(T_s) + H(T_s) + LE(T_s) - G_g(T_s) - G_s(T_s) = 0
```
where $G_g$ is the conductive heat flux over the snow-free fraction only (ground-atmosphere interface) and $G_s$ is the conductive heat flux for the snow-covered fraction. Note that many models with simplified snow schemes instead directly add a snow-melt term to the SEB, but this assumes that the snowpack does not store any energy, which is wrong on hourly time scales and even longer for deep snowpacks. So I think we'd like to avoid this.

In the above formulation of the SEB, the skin temperature would still be solved for using an aggregated albedo in the $R_\text{net}$ term which might be another source of problems since it results in the snow-covered fraction seeing a potentially unrealistic radiation budget. I will investigate whether or not the above partitioning of the ground heat flux actually helps or whether it makes more sense to just move to a tiling-based approach.

bgroenks96:
Fixes #177

The following is a summary of the problems and fixes, generated by Claude :robot:

## Summary
Northern-latitude winter SWE was staying near-zero (<0.001 m) despite persistent sub-zero forcing, with snow that did accumulate sporadically appearing during summer months too. Investigation traced this to four independent bugs spanning the surface energy balance (SEB), snow energy closure, and snow-top conduction discretization. All four are fixed and verified, both in unit/regression tests and in real coupled SpeedyWeather+Terrarium runs (3-month and full 1-year).

## Problems identified
Unweighted bulk SEB flux forcing the snowpack directly. The single, grid-cell-blended skin-temperature solve produced one bulk ground heat flux G (mixing snow/bare-ground temperature, conductivity, and albedo), which was then applied unmediated — with zero f_snow weighting — as the snowpack's own top-of-pack forcing. A cell with 10% snow cover received the full cell-scale flux, driving thin/patchy snow straight through its melting point every step.
Energy-tendency gate suppressing cold-snow advection on the accumulation-onset step. compute_snow_energy_tendency's (W > 0) gate zeroed the entire tendency — including the precipitation-advection term — on the very first step any snow landed on bare ground, so fresh snow entered with zero energy, which the FreeWater closure reads as instantaneously fully melted.
Liquid-fraction/temperature closures sharing an inappropriate floor. energy_to_temperature! computed both snow temperature and liquid fraction from the same min_conduction_thickness-floored volumetric energy. The floor is needed to keep the temperature calculation from blowing up as depth → 0, but it diluted a genuinely cold, thin layer's energy density toward the melting threshold for liquid fraction, causing spurious full melt.
Explicit-Euler numerical instability in the snow-top conductive flux. At the default 5mm min_conduction_thickness, the implied relaxation timescale (~30s) was ~9x shorter than the model's 300s timestep, so snow temperature oscillated wildly (~0°C ↔ ~-70°C) every step. Landing on the warm side snapped snow_liquid_fraction to 1.0 and fired the Darcy meltwater term at its fully-saturated rate, wiping out accumulated SWE in a single step — confirmed via direct per-step tracing in a coupled year-long run (outputs/run_0005).

## Summary of changes
SEB partitioning: skin_temperature.jl now solves the residual with two unblended, area-weighted conductances — G (skin → bare ground) and S (skin → snow top) — instead of one blended term. ground_heat_flux becomes an explicit conductive flux at the converged Ts, and a new surface_heat_flux field (properly f_snow-weighted) drives the snowpack's own tendency, replacing the old unweighted alias.
Energy tendency gate: excludes the cold-snow precipitation-advection term Q_prcp from the (W > 0) gate — it's a pure source term, always well-defined.
Liquid fraction vs. temperature: energy_to_temperature! now computes liquid fraction from the unfloored, depth-integrated energy while temperature keeps the floored path — decoupling two consumers of the same quantity with different numerical needs.
Grid-derived conduction floor: min_conduction_thickness is no longer a hand-tuned struct constant; it's now min_snow_conduction_thickness, computed as half the uppermost soil grid cell's thickness, tying the floor to the grid's own resolution.
Adaptive timestepping: added cell_diffusion_timescale for snow (mirroring the existing soil implementation), wired into LandModel as min(soil_τ, snow_τ) so TimeStepWizard can shrink Δt when the snow conduction stability limit genuinely requires it, rather than relying solely on a fixed floor.
Basal heat flux: compute_snow_basal_heat_flux moved to snow_interfaces.jl and rewritten as a proper series-resistance formula (Q_base = (T_soil − T_snow) / (Δz_soil/2κ_soil + d_snow/2κ_snow)), correctly accounting for the soil's own half-cell resistance instead of assuming it negligible.

## Testing
Full regression suite: 7902/7902 pass (test/snow/*, test/coupled_models/land_model_tests.jl, test/surface/*, including the 7776-case LandModel stress sweep).
Rewritten basal-flux and gate/liquid-fraction regression tests verified against independently reimplemented closed-form formulas.
Differentiability: test/differentiability/snow_model_diff.jl rewritten to differentiate the new field-based compute_snow_basal_heat_flux via Enzyme (Duplicated(state, dstate)), fixing a stale-signature CI failure.
Clean local doc build (docs/make.jl --local --draft), including updated ImplicitSkinTemperature docstring/process page.
Coupled model verification: 3-month (outputs/run_0006) and full 1-year (outputs/run_0008) SpeedyWeather+Terrarium runs analyzed via xarray. Representative seasonal-snow sites (S. Siberia/Mongolia, US Great Plains, Scandinavia) show a physically correct annual cycle — winter accumulation, spring peak, full summer melt-out tracking skin temperature crossing 0°C, autumn re-accumulation — with zero single-step drainage-wipeout events anywhere on the globe over the full year.

## Known limitations
Still solves a single skin temperature/albedo (no full tiling of snow-covered vs. bare-ground sub-fractions); turbulent fluxes remain unpartitioned. Flagged as future work if residual error proves significant.
ground_thermal_interface still uses a constant soil thermal conductivity rather than the soil model's own state-dependent value — a longstanding simplification, scoped as separate future work.
(Full investigation history, all revisions, and design rationale: docs/dev/2026-08/2026-08-27_PLAN_01_snow_seb_partitioning_investigation.md)

### DIFF
--- a/src/diagnostics/surface_fluxes.jl
+++ b/src/diagnostics/surface_fluxes.jl
@@ -16,3 +16,55 @@ function diagnose_skin_temperature_residual(
         return Ts_prev - Ts_implicit
     end
 end
+
+"""
+    $TYPEDSIGNATURES
+
+Return a `KernelFunctionOperation` that computes the kinematic surface humidity flux [m/s] over all grid cells.
+"""
+diagnose_surface_humidity_flux(state, model::LandModel) =
+    diagnose_surface_humidity_flux(state, model, get_evapotranspiration(model.surface_hydrology), model.constants, model.atmosphere)
+
+function diagnose_surface_humidity_flux(
+        state,
+        model::AbstractModel,
+        evapotranspiration::AbstractEvapotranspiration,
+        constants::PhysicalConstants,
+        atmos::AbstractAtmosphere
+    )
+    return kernel_operation2D(compute_surface_humidity_flux, state, model.grid, evapotranspiration, constants, atmos)
+end
+
+"""
+    $TYPEDSIGNATURES
+
+Return a `KernelFunctionOperation` that computes the evaporation flux [m/s liquid water] over all grid cells.
+The kinematic humidity flux is rescaled by the air-to-water density ratio and weighted by the snow-free area fraction.
+"""
+diagnose_evaporation_flux(state, model::LandModel) =
+    diagnose_evaporation_flux(state, model, get_evapotranspiration(model.surface_hydrology), model.constants, model.atmosphere, model.snow)
+
+function diagnose_evaporation_flux(
+        state,
+        model::AbstractModel,
+        evapotranspiration::AbstractEvapotranspiration,
+        constants::PhysicalConstants,
+        atmos::AbstractAtmosphere,
+        snow::Optional{AbstractSnow} = nothing
+    )
+    return kernel_operation2D(evaporation_flux, state, model.grid, evapotranspiration, constants, atmos, snow)
+end
+
+function evaporation_flux(
+        i, j, grid, fields,
+        evapotranspiration::AbstractEvapotranspiration,
+        constants::PhysicalConstants,
+        atmos::AbstractAtmosphere,
+        snow::Optional{AbstractSnow}
+    )
+    Qh = compute_surface_humidity_flux(i, j, grid, fields, evapotranspiration, constants, atmos)
+    ρ_a = air_density(i, j, grid, fields, atmos, constants)
+    ρ_w = constants.material.density_water
+    f_snow = snow_cover_fraction(i, j, grid, fields, snow)
+    return (1 - f_snow) * Qh * ρ_a / ρ_w
+end
--- a/src/diagnostics/surface_fluxes.jl
+++ b/src/diagnostics/surface_fluxes.jl
@@ -52,19 +52,18 @@ function diagnose_evaporation_flux(
         atmos::AbstractAtmosphere,
         snow::Optional{AbstractSnow} = nothing
     )
-    return kernel_operation2D(evaporation_flux, state, model.grid, evapotranspiration, constants, atmos, snow)
-end
-
-function evaporation_flux(
-        i, j, grid, fields,
-        evapotranspiration::AbstractEvapotranspiration,
-        constants::PhysicalConstants,
-        atmos::AbstractAtmosphere,
-        snow::Optional{AbstractSnow}
-    )
-    Qh = compute_surface_humidity_flux(i, j, grid, fields, evapotranspiration, constants, atmos)
-    ρ_a = air_density(i, j, grid, fields, atmos, constants)
-    ρ_w = constants.material.density_water
-    f_snow = snow_cover_fraction(i, j, grid, fields, snow)
-    return (1 - f_snow) * Qh * ρ_a / ρ_w
+    function surface_evapotranspiration_flux(
+            i, j, grid, fields,
+            evapotranspiration::AbstractEvapotranspiration,
+            constants::PhysicalConstants,
+            atmos::AbstractAtmosphere,
+            snow::Optional{AbstractSnow}
+        )
+        Qh = compute_surface_humidity_flux(i, j, grid, fields, evapotranspiration, constants, atmos)
+        ρ_a = air_density(i, j, grid, fields, atmos, constants)
+        ρ_w = constants.material.density_water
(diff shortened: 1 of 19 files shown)


src/diagnostics/surface_fluxes.jl
```diff
@@ -16,3 +16,55 @@ function diagnose_skin_temperature_residual(
...
+        model::AbstractModel,
+        evapotranspiration::AbstractEvapotranspiration,
+        constants::PhysicalConstants,
+        atmos::AbstractAtmosphere,
+        snow::Optional{AbstractSnow} = nothing
+    )
+    return kernel_operation2D(evaporation_flux, state, model.grid, evapotranspiration, constants, atmos, snow)
+end
+
+function evaporation_flux(
```
maximilian-gelbrecht: What's the difference between this function and the other `compute_evapotranspiration_fluxes!` that we already have for `BareGroundEvaporation{NF}` and `PALADYNCanopyEvapotranspiration{NF}`?

maximilian-gelbrecht (commented):
So far I just have one comment. I am afraid I just don't have the time at the moment to look into this enough that I understand the physics implemented here. Because without looking deeper into it, I don't understand what's going on here.

src/diagnostics/surface_fluxes.jl
bgroenks96: This is a diagnostic. It doesn't overwrite any state; it's meant to reproduce a quantity that is not stored directly in any `Field` for debugging and (later) output.

I think the reason that I reproduced some of the code here was so that it would work for both ET implementations without assuming the presence of, e.g. `transpiration` in `fields`, which would fail for bare ground.

bgroenks96: Probably I should move it inside of the `diagnose_evaporation_flux` function so that it doesn't show up in the package namespace.

bgroenks96: Changed in [e17c15b](https://github.com/NumericalEarth/Terrarium.jl/pull/184/commits/e17c15b4f6d29eec5d3d126bff211f99978212f2).

maximilian-gelbrecht (approved):
I'll approve it because it looks sound to me and the results look good. 

If you want me to look at something particular still, point me directly to it. Otherwise, as I wrote, it's difficult for me to follow the implementation here without spending a lot of time on reading up about the physics involved.

PR #188, opened 2026-08-31
Move surface excess water pool to runoff process

### LINKED ISSUE #189: Richards `LandModel` water budget is non-conservative across ET, runoff and surface storage
## Summary

The standard `LandModel → SoilEnergyWaterCarbon → SoilHydrology{RichardsEq}` pathway does not conserve water across soil moisture, evapotranspiration, surface storage and runoff.

I think this happens for five different (but related) reasons:

|  # | Problem                                           | Source                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                         |
| -: | ------------------------------------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------ |
|  1 | `LandModel` does not pass its evapotranspiration or runoff schemes into the Richards tendency calculation   | [`LandModel.compute_tendencies!`, lines 116–123](https://github.com/NumericalEarth/Terrarium.jl/blob/v0.1.6/src/models/coupled/land_model.jl#L116-L123); [`SoilEnergyWaterCarbon.compute_tendencies!`, lines 77–90](https://github.com/NumericalEarth/Terrarium.jl/blob/v0.1.6/src/processes/soil/soil_coupled.jl#L77-L90); [Richards arguments default to `nothing`, lines 62–79](https://github.com/NumericalEarth/Terrarium.jl/blob/v0.1.6/src/processes/soil/hydrology/soil_hydrology_rre.jl#L62-L79); [`SurfaceHydrology.compute_tendencies!` only advances canopy interception, current `main` lines 50–58](https://github.com/NumericalEarth/Terrarium.jl/blob/main/src/processes/surface/surface_hydrology.jl#L50-L58) |
|  2 | An `XYZ` Richards kernel writes the vertically reduced `XY` surface-water tendency from every soil layer | [`surface_excess_water` declared as `XY`, lines 16–23](https://github.com/NumericalEarth/Terrarium.jl/blob/v0.1.6/src/processes/soil/hydrology/soil_hydrology_rre.jl#L16-L23); [combined `XYZ` kernel writes both tendencies, lines 165–176](https://github.com/NumericalEarth/Terrarium.jl/blob/v0.1.6/src/processes/soil/hydrology/soil_hydrology_rre.jl#L165-L176)                                                                                                                                                                                                                                                                                                                                                          |
|  3 | Porosity is not included when redistributing water through saturated soil            | [`adjust_saturation_profile!`, lines 151–179](https://github.com/NumericalEarth/Terrarium.jl/blob/v0.1.6/src/processes/soil/hydrology/soil_hydrology.jl#L151-L179), particularly the [top-layer transfer at lines 176–179](https://github.com/NumericalEarth/Terrarium.jl/blob/v0.1.6/src/processes/soil/hydrology/soil_hydrology.jl#L176-L179)                                                                                                                                                                                                                                                                                                                                                                                |
|  4 | Drainage has the wrong reservoir sign and the limiter has the wrong units | [`compute_surface_drainage`, current `main` lines 20–31](https://github.com/NumericalEarth/Terrarium.jl/blob/main/src/processes/surface/runoff/direct_surface_runoff.jl#L20-L31); [`compute_surface_excess_water_tendency`, lines 221–243](https://github.com/NumericalEarth/Terrarium.jl/blob/v0.1.6/src/processes/soil/hydrology/soil_hydrology.jl#L221-L243)                                                                                                                                                                                                                                                                                                                                                                |
|  5 | Bottom-layer clamp creates water                  | [`adjust_saturation_profile!`, lines 181–183](https://github.com/NumericalEarth/Terrarium.jl/blob/v0.1.6/src/processes/soil/hydrology/soil_hydrology.jl#L181-L183)                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                             |

For a bare-soil column,

$$
W=\sum_k\phi_k S_k\Delta z_k+S_{\rm surface},
$$

where $S_k$ is relative saturation, $\phi_k$ is porosity and $S_{\rm surface}$ is surface-water depth.

Ignoring lateral transport,

$$
\Delta W=\int\left(P-E_{\rm ground}-T-R-q_{\rm bottom}\right)\,dt.
$$

## Details

### 1. Evapotranspiration and runoff are not handed over

`LandModel.compute_tendencies!` calls the surface-hydrology and soil tendencies separately:

```julia
compute_tendencies!(state, grid, model.surface_hydrology)
compute_tendencies!(state, grid, model.soil, model.constants)
```

The soil wrapper then calls Richards *without* the surface processes. The Richards method consequently receives:

```julia
evtr = nothing
runoff = nothing
```

As a result
* ground evaporation and transpiration do not remove water from the soil
* diagnosed drainage/runoff can export water without reducing surface storage
* the standard coupled pathway cannot close its water budget

### 2. An `XYZ` kernel writes an `XY` field

`surface_excess_water` is an `XY` prognostic field, but the combined Richards kernel uses `XYZ` and calls its tendency calculation for every `(i,j,k)` work item.

The saturation and surface-water calculations should use separate kernels:

* `XYZ` for soil saturation;
* `XY` for surface water and drainage.

I think as written the current version risks overwriting drainage tendencies incorrectly.

### 3. Closure transfers omit porosity

The top-layer correction currently applies:

```julia
surface_excess_water[i, j, 1] += excess_sat * Δz_top
```
Because `excess_sat` is a change in relative saturation, the physical water removed from the soil is

$$
\phi_\mathrm{top} \Delta S \Delta z_\mathrm{top}.
$$

With $\phi=0.49$, the surface reservoir receives approximately twice the water removed from the soil.

Transfers between layers also require the donor-to-receiver porosity ratio when porosity varies:

$$
\Delta S_{k+1}=\Delta S_k
\frac{\phi_k\Delta z_k}
     {\phi_{k+1}\Delta z_{k+1}}.
$$

### 4. Surface drainage has the wrong sign and limiter

`DirectSurfaceRunoff` defines positive drainage as

$$
D=S_{\rm surface}/\tau_r
$$

and includes it in exported runoff. However, `compute_surface_excess_water_tendency!` adds the same positive value to the surface reservoir.

Drainage should instead give

$$
\frac{\partial S_{\rm surface}}{\partial t}=-D.
$$

The current limiter,

```julia
min(∂S∂t, S)
```
also compares a rate in m s⁻¹ against a depth in m and so isn't timestep aware.

### 5. Bottom-layer closure creates water

The saturation closure finishes with:

```julia
sat[i, j, 1] = max(sat[i, j, 1], sat_min)
```

## Fixes
I've tried to fix these in [ReadyESM](https://github.com/kieranmrhunt/ReadyESM).

1. Some edits to `LandModel.compute_tendencies!` and passes the live evapotranspiration and runoff processes instead of allowing both to default to `nothing`. [Implementation](https://github.com/kieranmrhunt/ReadyESM/blob/main/src/terrarium_gpu_coupling.jl#L638-L675)
2.  Soil saturation is calculated with an `XYZ` kernel, while surface-water drainage is calculated once with an `XY` kernel. [Implementation](https://github.com/kieranmrhunt/ReadyESM/blob/main/src/terrarium_gpu_coupling.jl#L481-L603)
3. Runoff exports $S/\tau_r$, so the surface reservoir receives:

   ```math
   \frac{\mathrm dS}{\mathrm dt}=-\frac{S}{\tau_r}.
   ```
   [Implementation](https://github.com/kieranmrhunt/ReadyESM/blob/main/src/terrarium_gpu_coupling.jl#L454-L479)
4. The physical infiltration flux is divided by soil porosity before becoming a relative-saturation boundary flux:

   ```math
   F_s=-\frac{I}{\phi}.
   ```
   [Implementation](https://github.com/kieranmrhunt/ReadyESM/blob/main/src/terrarium_gpu_coupling.jl#L334-L452)

I also stop a duplicated Darcy flux at the upper boundary [Implementation](https://github.com/kieranmrhunt/ReadyESM/blob/main/src/terrarium_gpu_coupling.jl#L604-L636)




## Tests

In a coupled ERA5-land experiment:

* the worst land column gained approximately **0.259 m of water in one day**;
* **0.257 m** was retained at the surface;
* the resulting runoff reduced coastal salinity to **5.607 psu on day 1** and negative values after day 2.

A simple six-hour column experiment gave me:

| Diagnostic         |      Original | Repaired |
| ------------------ | ------------: | ----------------------: |
| Storage change     |     +5.697 mm |             +1.30683 mm |
| Net boundary input |     −0.057 mm |             +1.30607 mm |
| Budget residual    | **+5.754 mm** |        **+0.000761 mm** |
| Peak surface water |      4.058 mm |                0.481 mm |

After separating the kernels, four independent GPU drainage trajectories all produced `0.0186241399 m`, compared with the expected `0.0186241660 m`.

bgroenks96:
The `surface_excess_water` was previously handled by soil hydrology since it was originally needed there. However, it makes more sense being owned by `AbstractSurfaceRunoff` implementations.

This is a prerequisite change for some work on refactoring soil, surface hydrology, and vegetation processes to use fused kernels.

See the plan doc.

Edit: This PR also happens to fix the bugs mentioned in #189.

Closes #189


src/models/coupled/land_model.jl
```diff
@@ -145,14 +145,16 @@ end
 
 function closure!(state, model::LandModel)
     grid = get_grid(model)
-    closure!(state, grid, model.soil, model.constants)
+    # Pass the surface runoff process so excess water removed from an oversaturated soil surface is
+    # routed into the runoff-owned `surface_excess_water` pool (which the runoff tendency then drains).
+    closure!(state, grid, model.soil, model.constants, model.surface_hydrology.surface_runoff)
```
maximilian-gelbrecht: Is there always a `surface_runoff` for every `LandModel` you can (in theory) construct? 

Maybe a good place for a getter function with `nothing` fallback.

maximilian-gelbrecht (approved):
Yeah, looks good to me. Given the very general plan where/how would you continue? Did you benchmark in some way that this is beneficial not just to fixing the existing issues remarked in #189? Yesterday you said it's actually slower, or was this something else?

src/models/coupled/land_model.jl
bgroenks96: Fixed in [943d5ac](https://github.com/NumericalEarth/Terrarium.jl/pull/188/commits/943d5ac7289820933e76339efa77f43a8f40ffb3)

PR #192, opened 2026-09-16
Implement fused surface hydrology `compute_auxiliary!`

bgroenks96:
This PR adds kernel fusion for surface hydrology and vegetation. Description below will be updated after those changes are added.

*The following is generated by an LLM :robot:*

## Fuse `SurfaceHydrology.compute_auxiliary!` into a single launch

The coupled `SurfaceHydrology` auxiliary pass fanned out to three XY launches per step
(canopy interception → evapotranspiration → surface runoff). This PR collapses them into one
fused `compute_auxiliary_kernel!` dispatching on `SurfaceHydrology`, calling the per-cell
mutating variants in dependency order so each stage reads the previous stage's outputs from
the aliased `fields`. This follows the structure established by `SurfaceEnergyBalance` and the
fusion rule from the implementation plan: *fuse a dependency chain, not independent outputs*
(see `docs/dev/2026-08/2026-08-31-PLAN_fused_kernels.md`, rev 6–7).

### Changes

- **`surface_hydrology.jl`** — fused host method + `@kernel`; `out` is the union of the three
  sub-processes' auxiliaries, with lazy `FunctionField`s (e.g. `NoCanopyInterception`'s
  `rainfall_ground` passthrough) filtered out since no kernel writes them.
- **`compute_evapotranspiration_auxiliary!`** — new uniform-signature kernel function defined
  for both ET schemes (PALADYN canopy + bare ground); each scheme's standalone
  `compute_auxiliary_kernel!` is now a thin wrapper over it, so the fused and standalone paths
  cannot drift.
- **`canopy_interception.jl`** — `NoCanopyInterception` gains a no-op
  `compute_canopy_auxiliary!` so the fused kernel dispatches uniformly.

### Bug fix: canopy ET never saw `snow`

The per-process canopy-ET `compute_auxiliary!` accepted `snow` but its `launch!` forwarded only
up to `vegetation`, so `snow` fell into `args...` and the kernel defaulted to `snow = nothing` —
canopy ground/canopy evaporation and transpiration were **never scaled by the snow-free
fraction**, unlike bare-ground ET. The fused kernel passes `snow` to both schemes and the
standalone PALADYN host method is fixed to match. **This is a deliberate numerics change** for
the vegetated-under-snow configuration (ET is now suppressed over snow-covered ground, as it
should be).

### Tests

`test/surface/hydrology/integration_tests.jl`:
(i) fused vs per-process fan-out over all 12 surface-hydrology auxiliaries, vegetated + snow,
to machine precision; (ii) canopy ET fluxes == (1 − f_snow) × snow-free fluxes.
Existing `test/surface/*` and `test/coupled_models/land_model_tests.jl` green.

### Benchmarks

Micro (isolated `compute_auxiliary!`, min-of-5, CPU):

| columns | baseline (ms/step) | fused (ms/step) | Δ |
|--------:|-------------------:|----------------:|---:|
| 512     | 0.220 | 0.257 | −14.5% |
| 4,608   | 0.391 | 0.381 | +2.6%  |
| 18,432  | 0.878 | 0.670 | +30.9% |
| 73,728  | 2.916 | 2.024 | +44.0% |

The three passes' redundant re-reads of the shared state collapse to one, which dominates at
production resolutions. The −14.5% at 512 columns is 0.037 ms/step — negligible against a full
land-model step. GPU and full-model-step numbers pending (exclusive-node A/B running).


src/processes/vegetation/vegetation_carbon_cycle.jl
```diff
@@ -87,36 +87,73 @@ function compute_auxiliary!(
...
-    compute_auxiliary!(state, grid, veg.autotrophic_respiration, veg.carbon_dynamics, veg.phenology, veg.traits, atmos)
+    # The carbon-cycle chain — carbon dynamics → phenology → photosynthesis → stomatal conductance →
+    # autotrophic respiration — is fused into a single launch. Each stage reads the previous stage's
+    # output within a cell, so the dependency chain resolves without returning to the host.
+    carbon_dynamics = veg.carbon_dynamics
+    phenology = veg.phenology
+    photosynthesis = veg.photosynthesis
+    stomatal_conductance = veg.stomatal_conductance
+    autotrophic_respiration = veg.autotrophic_respiration
+    out = filter(
```
maximilian-gelbrecht: Why is this needed? I don't quite understand.

maximilian-gelbrecht (commented):
Looks pretty straight forward to me, just one question about the output field assignment (see comments)

src/processes/vegetation/vegetation_carbon_cycle.jl
bgroenks96: It's not, looks like a Claude artifact that I missed. Will fix.

Issue #182, opened 2026-08-26
Rename grid types and add multi-domain support with `VarDomain`


bgroenks96:
Rename existing grid types (e.g., `ColumnGrid` → `ColumnLandGrid` and `ColumnRingGrid` → `GlobalColumnLandGrid` or `RingColumnLandGrid`) and add separate domain grids for Ground, Snow, and Canopy. Introduce `VarDomain` type for the variable system to tag variables with their domain.

Detailed design to follow.

PR #193, opened 2026-09-18
Refactor grid types to support all Oceananigans grids as well as snow/canopy vertical domains

### LINKED ISSUE #183: Implement generic `LandGrid` supporting all Oceananigans grids
`LandGrid{NF, TX, TY, TZ, Arch, G<:AbstractGrid}` will wrap any valid `Oceananigans` grid (ironically even another `AbstractLandGrid`... but maybe we should explicitly disallow that) and per #182 three domain grids of type `G` with distinct vertical discretizations. Constructor builds each domain grid from shared horizontal discretization + per-domain `AbstractVerticalSpacing`. Domain access via `get_domain_grid(grid, Val(:ground))`. Update `Field` and `launch!` to dispatch on domain grid.

bgroenks96:
This PR closes the gap between Terrarium's grid types and Oceananigans'. Any `AbstractGrid` can now be
handed to a model constructor, and the per-domain (ground, snow, canopy) structure a land model
needs is built internally rather than being baked into the grid the user writes down.

Implements Phases 1 to 4 of
[`docs/dev/2026-08/2026-08-25-PLAN_oceananigans_grid_integration.md`](docs/dev/2026-08/2026-08-25-PLAN_oceananigans_grid_integration.md),
which carries the full design and its revision history. The `VarDomain` half of Phase 3 is
deliberately **not** in this PR; see [Deferred to a follow-up PR](#deferred-to-a-follow-up-pr).

## What changes

**`ColumnGrid` is a type alias, not a wrapper.** It is now
`RectilinearGrid{NF, Periodic, Flat, Bounded, ...}` plus constructors, so a column grid simply *is*
an Oceananigans grid. `AbstractColumnGrid` and the `WrappedGrid` union are gone.

**`ColumnRingGrid` stays a struct** because it carries the ring grid and land mask, but it subtypes
`AbstractGrid` directly and implements the Oceananigans grid interface itself.

**One `AbstractLandGrid` implementation, `LandGrid`,** wrapping separate ground, snow, and canopy
grids derived from the same underlying `AbstractGrid`. Model structs are unchanged: `grid` is still
the first field and still annotated `AbstractLandGrid`. Outer constructors accept a bare
`AbstractGrid` and build the `LandGrid` through `create_land_grid(grid, soil, snow, vegetation)`,
which currently returns a ground-only grid, correct for every implemented (0D) snow and vegetation
scheme. `create_land_grid` is idempotent on a `LandGrid`.

**`get_field_grid` is removed.** Every grid is an `AbstractGrid` now, so code either uses the grid it
was given or, where the vertical dimension is genuinely ambiguous (soil energy and hydrology kernels,
root distribution, plant-available water), asks for `ground_domain(grid)` explicitly.

**Vertical discretizations come from Oceananigans.** `AbstractVerticalSpacing`, `PrescribedSpacing`
and `get_spacing` are deleted. A uniform or prescribed column is given directly as a range or vector
of cell interfaces. `UniformSpacing(; Δz, N)` survives as a convenience returning such a range, and
`ExponentialSpacing(; Δz_min, Δz_max, N)` as a function returning an Oceananigans
`ExponentialDiscretization`: the two describe the same geometric family, and the land-model
parameterization converts in closed form via `ρ = (Δz_max/Δz_min)^(1/(N-1))`,
`D = Δz_min (ρᴺ-1)/(ρ-1)`, `scale = D/(N log ρ)`. Layer thicknesses agree to floating point; the only
visible change is the loss of the old significant-digit rounding, worth about 0.4% on a default
profile.

**`launch!` dispatches on the workspec.** Terrarium's convenience `launch!` previously dispatched on
the grid, which would have meant type piracy once it had to accept any `AbstractGrid`. It now
dispatches on the Terrarium-owned `XY`/`XYZ` workspec instead, and `get_workspec` is gone. The
`compute_z_bcs!` indirection is deleted for the same reason: it existed only to turn a grid into an
architecture, and its call sites now pass `architecture(grid)` directly.

## Making `ColumnRingGrid` a first-class Oceananigans grid

`ColumnRingGrid` is meant to be usable anywhere a `RectilinearGrid` is. Oceananigans does not define
what that requires: `src/Grids/abstract_grid.jl` declares the supertype and a handful of methods, and
the rest is found by running into a `MethodError`.

Three gaps surfaced here, none of them mentioned in `abstract_grid.jl`:

| Method | Needed by | Exported by `Oceananigans.Grids` |
|---|---|---|
| `ξname`, `ηname`, `rname` | `set!(::Field, ::Function)`, to name the coordinate arguments | no |
| `ξnode`, `ηnode`, `rnode` | `set!(::Field, ::Function)`, to evaluate coordinates per node | yes |
| `serializeproperty!` | `JLD2Writer` output and `FieldTimeSeries` read-back | no |

The serialization one is a genuine incompatibility rather than a missing forward: a `ColumnRingGrid`
cannot round-trip through JLD2, because its `mask` is a `RingGrids.Field{Bool}` backed by a
`Vector{Bool}` that JLD2 reads back as a `BitVector`, failing a `typeassert` against the grid's own
`Mask` type parameter. We serialize the wrapped `RectilinearGrid` instead and reattach the horizontal
geometry on load by passing the live grid to `RingGrids.Field`.

Oceananigans also reads grid geometry as *struct fields* (`Nx`, `Hx`, `xᶜᵃᵃ`, `z`, ...) rather than
through accessors, so a wrapper grid cannot be implemented by defining methods alone and must forward
`getproperty`. This is worth raising upstream: a conformance test suite that a candidate grid could be
run against would have caught all three of these at once.

## Breaking changes

No backward compatibility is kept. For user code:

| Before | After |
|---|---|
| `get_field_grid(grid)` | `ground_domain(grid)`, or just use `grid` |
| `UniformSpacing(Δz = 0.1, N = 100)` as a spacing object | same call, now returns a range of interfaces |
| `PrescribedSpacing(z)` | pass `z` directly |
| `ExponentialSpacing(...)` | same call, now returns an `ExponentialDiscretization` |
| `get_workspec(...)` | removed; `launch!(grid, XY(), ...)` unchanged at call sites |

Newly exported: `LandGrid`, `create_land_grid`, `ground_domain`, `snow_domain`, `canopy_domain`,
`num_layers`. Removed: `get_field_grid`, `PrescribedSpacing`.

## Testing

- Full suite green on Julia 1.10 and 1.12.
- Reactant correctness suite green in CI (Julia 1.12): CPU-vs-Reactant comparison across five model
  configurations, plus the Reactant + Enzyme autodiff tests.
- Enzyme suite green. Worth noting that four Enzyme tests failed on `main` before this branch and
  pass now: removing the wrapper indirection let Enzyme's type analysis through.
- Documentation builds, with all examples executed and doctests passing.
- Tests construct models from bare `AbstractGrid`s, as users do, rather than wrapping grids in a
  `LandGrid` explicitly. `test/grids.jl` is the exception, since `LandGrid` is what it tests.

## Deferred to a follow-up PR

**`VarDomain` tagging (Phase 3, step 4) is not in this PR.** It is in progress on `bg/var-domains`.

This PR gives `LandGrid` somewhere to *put* a snow or canopy discretization; it does not give
variables a way to *say* which domain they live on. Every variable is therefore allocated on the
ground domain, which is correct today because no implemented scheme resolves snow or canopy
vertically, so the other two domains are always `nothing`.

The follow-up adds a `VarDomain` tag (`Ground`, `Snow`, `Canopy`), pairs it with the existing
`VarDims` in a `VarLocation` carried by every `Variable`, and routes `Field` allocation and `launch!`
through the domain grid it selects. It also introduces a `Coordinate` location so a variable can be
declared at a single point along an axis, which is what boundary fluxes at the `Top` or `Bottom` of a
domain need. That work touches every `variables()` declaration in the package, so it is kept separate
to keep this diff reviewable.

Also deferred with it: the per-component `ground_discretization`, `snow_discretization` and
`canopy_discretization` methods. Building a domain grid from a spacing needs a grid-type-specific
"same horizontal discretization, new vertical discretization" operation, and nothing needs it until a
vertically resolved scheme exists.

Closes #183

Partially addresses #178 and #182 

CC @glwagner @xkykai


docs/src/introduction/numerical_core.md
```diff
@@ -7,16 +7,31 @@ CurrentModule = Terrarium
...
+Terrarium distinguishes between two layers of grid type.
 
-All Terrarium `AbstractLandGrid` implementations are required to implement the following methods:
+**Spatial discretizations** are ordinary Oceananigans [`AbstractGrid`](@extref Oceananigans.Grids.AbstractGrid)s and describe how physical space is divided into control volumes. Any Oceananigans grid can be used, and Terrarium additionally defines two of its own, both based on the Oceananigans [`RectilinearGrid`](@extref Oceananigans.Grids.RectilinearGrid):
+
+- [`ColumnGrid`](@ref) represents an unstructured collection of 1D vertical columns discretized along the `Z` axis with the `X` axis corresponding to a non-spatial index over each of those columns and the `Y` axis ignored (assigned a `Flat` grid "topology"). `ColumnGrid` is simply an *alias* for a `RectilinearGrid` with topology `(Periodic, Flat, Bounded)` rather than a distinct type, together with constructors which build one from a vertical discretization; a `ColumnGrid` is in every respect an ordinary `RectilinearGrid`.
```
olivierbonte: Why should the x-axis be periodic for the cases of non communicating columns?

src/grids/column_grid.jl
```diff
@@ -1,55 +1,46 @@
...
 
-Represents a set of laterally independent vertical columns with dimensions (x, y, z)
-where `x` is the column dimension, `y=1` is constant, and `z` is the vertical axis.
+Note that this is a plain spatial discretization, not a land grid; see [`LandGrid`](@ref) for the
+land model counterpart which collects the discretizations of all vertical domains.
 """
-struct ColumnGrid{NF, Arch, RectGrid <: Oceananigans.Grids.RectilinearGrid} <: AbstractColumnGrid{NF, Arch}
-    "Underlying Oceananigans rectilinear grid on which `Field`s are defined."
-    grid::RectGrid
+const ColumnGrid{NF, Arch} = RectilinearGrid{NF, Periodic, Flat, Bounded, CZ, FX, FY, VX, VY, Arch, SZ} where {CZ, FX, FY, VX, VY, SZ}
```
olivierbonte: If there is an explanation as to why X is periodic, would be good to add it to docstrings here

src/grids/grids.jl
```diff
@@ -1,43 +1,47 @@
...
+Base.summary(grid::AbstractLandGrid) = "$(nameof(typeof(grid))) with dimensions $(size(grid))"
+Base.size(grid::AbstractLandGrid) = size(ground_domain(grid))
 
-Architectures.architecture(grid::AbstractLandGrid) = architecture(get_field_grid(grid))
+Architectures.architecture(grid::AbstractLandGrid) = architecture(ground_domain(grid))
 Architectures.on_architecture(arch, grid::AbstractLandGrid) = adapt(array_type(arch), grid)
 
-Oceananigans.Grids.halo_size(grid::AbstractLandGrid) = Oceananigans.Grids.halo_size(get_field_grid(grid))
-Oceananigans.Grids.isrectilinear(grid::AbstractLandGrid) = Oceananigans.Grids.isrectilinear(get_field_grid(grid))
+Oceananigans.Grids.halo_size(grid::AbstractLandGrid) = halo_size(ground_domain(grid))
```
olivierbonte: For now this makes sense, as only the soil grid is a proper grid, but how is this supposed to behave when you e.g. have a multi layer canopy with an Oceananigans grid (and halo)?

olivierbonte: same goes for other method extensions on AbstractLandGrid

test/Project.toml
```diff
@@ -15,6 +15,7 @@ Rasters = "a3a2b9e3-a471-40c9-b274-f788e487c689"
 RingGrids = "d1845624-ad4f-453b-8ff4-a8db365bf3a7"
 SpecialFunctions = "276daf66-3868-5448-9aa4-cd146d93841b"
 Statistics = "10745b16-79ce-11e8-11f9-7d13ad32a3b2"
+Terrarium = "80418d68-07fa-499d-ae2b-c12e531f5cd8"
```
olivierbonte: Why does Terrarium need to become a dependency of the test env for this PR?

olivierbonte (commented):
Don't know that much about all the internal Oceananigans grids default, but looks logical to me. Only some minor questions.

docs/src/introduction/numerical_core.md
bgroenks96: It could also be `Bounded`; I think in a way they are both "wrong" and a symptom of abusing the Oceananigans grid abstractions. The point of using `Periodic` was that it makes the "boundary conditions" for the x-axis automatic (i.e. periodic), and thus prevents the user from erroneously setting BCs for the x dimension which wouldn't be used. If x is `Bounded`, then it would have BCs that could (theoretically) be configured, though again, they wouldn't actually do anything.

I would be fine with changing this to `Bounded` though, if you think it would prevent confusion.

src/grids/grids.jl
bgroenks96: The idea is that the `LandGrid` defaults to the ground because that is the one component of the land surface that is always there, unconditionally, in all model configurations. For grid operations on the snow/canopy, the calling code would need to explicitly retrieve the grid with `snow_domain` or `canopy_domain`.

test/Project.toml
bgroenks96: it doesn't, I don't know how/why that happened. Maybe Claude was lazy and added it. Will remove.

src/grids/column_grid.jl
bgroenks96: Done in [b024c10](https://github.com/NumericalEarth/Terrarium.jl/pull/193/commits/b024c108610a45cd92bb98c907501041d1d4e2fa)

docs/src/introduction/numerical_core.md
bgroenks96: Addressed in [b024c10](https://github.com/NumericalEarth/Terrarium.jl/pull/193/commits/b024c108610a45cd92bb98c907501041d1d4e2fa)

test/Project.toml
bgroenks96: Done in [0033251](https://github.com/NumericalEarth/Terrarium.jl/pull/193/commits/0033251d0db8ebb4efad3d6bfba92f7ff264384d)

docs/src/introduction/numerical_core.md
```diff
@@ -7,16 +7,31 @@ CurrentModule = Terrarium
...
+```julia
+grid = ColumnGrid(ExponentialSpacing(Δz_min = 0.05, Δz_max = 100.0, N = 50), 10)
+model = SoilModel(grid)
+get_grid(model)                     # a `LandGrid` wrapping `grid`
+ground_domain(get_grid(model))        # the underlying domain `grid`
+```
+
+A pre-built `LandGrid` may also be passed to a model constructor directly, in which case it is stored as-is. The domain grids are retrieved with [`ground_domain`](@ref), [`snow_domain`](@ref), and [`canopy_domain`](@ref).
+
+`LandGrid` and `ColumnRingGrid` are both thin wrappers around another grid, and each forwards the `AbstractGrid` interface to the underlying grid. This incldues `struct` fields such as `grid.Nx`, which Oceananigans accesses directly. A land grid is required to implement the following methods:
```
maximilian-gelbrecht:
```suggestion
`LandGrid` and `ColumnRingGrid` are both thin wrappers around another grid, and each forwards the `AbstractGrid` interface to the underlying grid. This includes `struct` fields such as `grid.Nx`, which Oceananigans accesses directly. A land grid is required to implement the following methods:
```

ext/TerrariumReactantExt/grids.jl
```diff
@@ -1,16 +1,4 @@
 # Grid transfers to/from the device.
-#
-# The generic land-grid `on_architecture` (src/grids/grids.jl) uses `adapt(array_type(arch), …)`,
```
maximilian-gelbrecht: This is removed, but the functionality still covered somewhere?

src/grids/land_grid.jl
```diff
@@ -0,0 +1,154 @@
...
+only in their vertical discretization. Domains which are not resolved vertically by the model
+(e.g. a single-layer snow scheme or a big-leaf canopy) are represented by `nothing`.
+
+`LandGrid` is the only implementation of [`AbstractLandGrid`](@ref) and is normally constructed
+implicitly by the model constructors via [`create_land_grid`](@ref) rather than directly by the user.
+
+Properties:
+$(TYPEDFIELDS)
+"""
+struct LandGrid{NF, TX, TY, TZ, Arch, GroundGrid, SnowGrid, CanopyGrid} <: AbstractLandGrid{NF, TX, TY, TZ, Arch}
```
maximilian-gelbrecht: We discussed this a few times before. I think just having the separate grids here is a pragmatically good decision.

maximilian-gelbrecht (commented):
I went over this just very quickly now. I think it looks good already. I have to see whether I have more time before next week

ext/TerrariumReactantExt/grids.jl
bgroenks96: I think it isn't needed because `RectilinearGrid` presumably already covers it?

maximilian-gelbrecht: Looking into it, I also think it's not needed anymore. The fact that the Reactant unit tests pass also shows that but the reason is that the old 

```Julia 
Architectures.on_architecture(arch, grid::AbstractLandGrid) = adapt(array_type(arch), grid)
```

was replaced with 

```Julia
function Architectures.on_architecture(arch::AbstractArchitecture, grid::LandGrid)
    return LandGrid(
        on_architecture(arch, ground_domain(grid)),
        on_architecture(arch, snow_domain(grid)),
        on_architecture(arch, canopy_domain(grid)),
    )
end
```

as the comment there in the deleted lines already said the `adapt` was the problem. So now this is done in a better way and the new extended version in the extension not necessary anymore

test/timestepping/imex.jl
```diff
@@ -69,7 +69,7 @@ module IMEXTestTypes
 end
 
 @testset "Single timestepper integrates all prognostic variables" begin
-    grid = ColumnGrid(CPU(), Float64, UniformSpacing(N = 1))
+    grid = ColumnGrid(CPU(), Float64, UniformSpacing(Δz = 0.1, N = 1))
```
maximilian-gelbrecht: Just a minor question does `Δz` not have default value anymore or why is it added everywhere explicitly?

src/grids/grids.jl
olivierbonte: Okay, defenitely fine for now then, as long as it is clear once there would be multiple options in the future

test/timestepping/imex.jl
bgroenks96: It does have a default value, I guess Claude just chose to be explicit everywhere, which I don't mind tbh.

